PR #7, opened 2026-01-15
Feat: SmartPicker

### LINKED ISSUE #55: Add smart picker to the web ui
In Nextcloud we have a feature called smart picker that is used in a lot of places to insert links or text. 

https://github.com/nextcloud/richdocuments/blob/main/src/mixins/pickLink.js#L8-L19 is how the smart picker can be triggered.

We would want to have a button in Euro-Office webapps that then can send a message to the nextcloud app (maybe through post messages that get passed between the iframe and the parent frame).

- [ ]  We should add an assistant and smart picker button in addition, aligning with @marcoambrosini on where to put that

rikled:
This is a draft of the SmartPicker feature. 

Todo:

- [x] Support for plain text insertions, currently only links supported. Seems like this requires extension of the sdkjs API
- [x] Check for lock conditions of the button
- [x] Extend feature to other editors

This PR addresses: https://github.com/Euro-Office/web-apps/issues/55.
depends on: https://github.com/Euro-Office/onlyoffice-nextcloud/pull/2, https://github.com/Euro-Office/sdkjs/pull/3

### DIFF
--- a/.github/workflows/ci-build-upload.yml
+++ b/.github/workflows/ci-build-upload.yml
@@ -119,7 +119,7 @@ jobs:
             env:
                 URI: ${{steps.tag-dir.outputs.URI}}
             run: |
-                aws cloudfront create-invalidation \
-                    --distribution-id ${{ secrets.AWS_DISTRIBUTION_ID }} \
-                    --paths \
-                    "${URI}/*"
+                API_STATUS=$(aws apigateway test-invoke-method --rest-api-id ${{ secrets.AWS_REST_API_ID }} --resource-id "${{ secrets.AWS_RESOURCE_ID }}" \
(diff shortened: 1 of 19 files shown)


.github/workflows/ci-build-upload.yml
```diff
@@ -119,7 +119,7 @@ jobs:
             env:
                 URI: ${{steps.tag-dir.outputs.URI}}
             run: |
-                aws cloudfront create-invalidation \
```
juliusknorr, line 122: I'd say we can drop this whole workflow for uploading on our CI or disable it by changing the triggers

juliusknorr: Ah i see that is from other commits that were made upstream (or pending in #3), nevermind then, unrelated to this change

juliusknorr:
Will close this as it was replaced by the follow up work of Christoph.

PR #9, opened 2026-01-23
:wrench: :lipstick: ⚗️ Shift from png to svg icons

moodyjmz:
Png sprite maps take up build time and the pngs are not needed in 2026. OO already supports SVGs, we just shift it to always use it and not process PNGs.

It does break with OO, but the changes are mostly build related and the html/js changes are minimal

More detail in this ticket: https://github.com/Euro-Office/fork/issues/37


moodyjmz:
Parking whilst I look at other more pressing issue, will pick up again

moodyjmz:
Is kind of working - had some issues with SVGs and I wanted to avoid multiple network calls and associated overhead.

I need to check it a bit, but is using SVGs, builds much faster.

Fun git conflicts

juliusknorr:
Works well, i found a small UI issue with the arrow next to some icons not being middle aligned, otherwise works nice and much smoother build times 🚀 

Screenshot 2026-03-18 at 16 16 27

juliusknorr (approved):

PR #24, opened 2026-03-25
fix: :bug: Ensure that icons for all toolbars are displayed

moodyjmz:
app index.deploy icon references were missing for everything apart from doc, this fixes this.


Aiiaiiio (approved):
💯

PR #48, opened 2026-04-10
feat: :art: Improve output of translation key comparison

moodyjmz:
The translations add lot of bunk to the build output and not particularly useful, this tidies it up a bit and also highlights where most keys are missing

image

### DIFF
--- a/translation/merge_and_check.py
+++ b/translation/merge_and_check.py
@@ -1,6 +1,7 @@
 import sys
 import json
 import os.path
+from pathlib import Path
 
 lost_key_count, sum_key_count = 0, 0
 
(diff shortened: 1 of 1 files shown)


translation/merge_and_check.py
```diff
@@ -51,27 +83,95 @@ def compareJsonInFolder(path):
...
-    print('compare path', cwd, '\n')
     with open(path, 'r', encoding="utf-8") as pf:
         master_dict = json.load(pf)
 
-        #for root, dirs, files in os.walk(os.path.dirname(path)):
-            #for f in files:
+        results = {}
         files = [f for f in os.listdir(cwd) if os.path.isfile(os.path.join(cwd, f))]
         for f in files:
             if f != 'en.json' and f[-5:] == '.json':
```
Aiiaiiio: It took me a minute to understand this part. Why not utilize pathlib? eg:
```suggestion
            lang = Path(f).stem
            if lang != 'en' and Path(f).suffix == '.json':
```

Aiiaiiio (approved):

PR #50, opened 2026-04-14
feat: add Ctrl+Alt+H keyboard shortcut to toggle highlight color

MarsEB71:
## Summary

- Adds `Ctrl+Alt+H` shortcut to the document editor toolbar
- First press applies the current highlight color to selected text
- Second press removes the highlight if the selection's color matches the current highlight color

## How it works

Reads the highlight state directly from `api.get_TextProps()` on each keypress rather than relying on cached button state. Uses `_setMarkerColor('transparent')` (which calls `SetMarkerFormat(true, false)`) to remove the highlight — not `SetMarkerFormat(false)` which only exits painting mode without removing existing highlights.

## Test plan

- [ ] Select text → `Ctrl+Alt+H` → text highlighted in yellow
- [ ] With highlighted text selected → `Ctrl+Alt+H` → highlight removed
- [ ] Change highlight color via toolbar → `Ctrl+Alt+H` → applies new color
- [ ] No text selected → `Ctrl+Alt+H` → no crash

Fixes https://github.com/Euro-Office/DocumentServer/issues/55

### DIFF
--- a/apps/documenteditor/main/app/controller/Toolbar.js
+++ b/apps/documenteditor/main/app/controller/Toolbar.js
@@ -374,6 +374,22 @@ define([
             toolbar.mnuMultiChangeLevel.menu.on('show:after',           _.bind(this.onChangeLevelShowAfter, this, 2));
             toolbar.mnuMultiChangeLevel.menu.on('item:click',           _.bind(this.onChangeLevelClick, this, 2));
             toolbar.btnHighlightColor.on('click',                       _.bind(this.onBtnHighlightColor, this));
+            (function(me) {
+                var highlightKeymap = {};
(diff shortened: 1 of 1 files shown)


moodyjmz:
Not sure if this gets added, I can't decide that unilaterally, however I would give some feedback generally

Why IIFE?
Why place the logic code into the init block?
Would it be better to have the logic as a named method alongside the other named methods?

I see you are using Claude. You can ask it to look at the existing patterns in the file and follow convention. :)

juliusknorr:
Hi @MarsEB71 Thanks a lot for your contribution again. Is there any feedback to the questions from James meanwhile?

moodyjmz:
I would close this and feel free to re-open in the future @MarsEB71

PR #51, opened 2026-04-21
Update Dockerfile to use relative paths for COPY commands

arikw:
The docker image build fails due to non-existing paths.
After the change, the `docker build -t webapps -f .docker/web-apps.bake.Dockerfile` command works as expected.


moodyjmz:
@rikled is this one for you?

rikled:
Thanks for the PR. :) The Dockerfiles are not supposed to be used standalone. They are consumed by the docker-bake.hcl files in the main repositories (DocumentServer and DesktopEditors). Changing the paths here would break these. All dockerfiles assume a directory layout like in DocumentServer/DesktopEditors.

I will close this for now, but feel free to make suggestions on how to improve :)

PR #56, opened 2026-05-05
fix: :bug: Fix show-password toggle in PasswordField

wapec:
- Input type was 'type' instead of 'text', making the reveal non-functional
- Toggle button was a non-interactive <span>; replaced with <button> for keyboard accessibility with aria-pressed and aria-label
- aria-label uses i18n translations (textShowPassword / textHidePassword)
- Added CSS reset for the new button element to preserve existing layout

## Test plan

  > Open a file in the mobile editor.
  > Navigate: **Settings → Protection → Protect Document → Set Password**.

  ### Show / hide toggle — visibility
  - [ ] iOS: eye icon is visible on both Password and Verify fields
  - [ ] Android: eye icon is visible on both Password and Verify fields
  - [ ] (Regression) Before this PR the icon was missing on iOS

  ### Show / hide toggle — behaviour
  - [ ] Default state: input is masked (`••••••`)
  - [ ] Type a password (e.g. `test123456`) — still masked
  - [ ] Tap the eye icon → password becomes plain text
  - [ ] (Regression) Before this PR tapping the icon did nothing
        (input had `type="type"`, an invalid value)
  - [ ] Icon swaps to “hide” (crossed-out eye) when text is shown
  - [ ] Tap again → masked again, icon returns to “show”
  - [ ] Password and Verify fields toggle independently

  ### Layout
  - [ ] iOS: eye icon is right-aligned, vertically centred
  - [ ] iOS: tap target is at least 44×44 px
  - [ ] Android: layout unchanged vs. previous behaviour

  ### Submit flow (sanity)
  - [ ] Empty fields → **Save** disabled
  - [ ] Matching passwords → **Save** enabled
  - [ ] Mismatched passwords → validation error shown
  - [ ] Make sure translations are used in the template
image

  ### Screenshots
Password field is shown with IconHide and Verify field with IconShow
  | iOS | Android |
  |---|---|
  | ios | android |


moodyjmz:
I like the removal of the more convoluted logic for the icon - when out of draft? @wapec

wapec:
thanks @moodyjmz . I have just cleaned up the commits and split into two changelog entries. PR is ready for review

juliusknorr:
Thanks for the pull request. Could you elaborate a bit more on those points from your pull request description?

> Toggle button was a non-interactive ; replaced with for keyboard accessibility with aria-pressed and aria-label
> aria-label uses i18n translations (textShowPassword / textHidePassword)
> Added CSS reset for the new button element to preserve existing layout

There seems to be no aria attributes/buttomn element and the span with onClick handler is not keyboard focussable.

Other then the difference between diff and description i think this is a good addition, so would approve that already.

juliusknorr (approved):

juliusknorr:
About the test plan, is this something you had verified during development?

moodyjmz:
**Accessibility fix from the description is still missing**

@juliusknorr already flagged this, but worth spelling out the full change needed to avoid a partial fix landing.

The `<span onClick>` at line 33 needs to become a native button:

```jsx
<button
    type="button"
    onClick={toggleShowPassword}
    aria-pressed={isShowPassword}
    aria-label={isShowPassword ? t('textHidePassword') : t('textShowPassword')}
    className='password-field__toggle'
>
```

A few things to land alongside this:

- `type="button"` is required — the component renders inside a Framework7 form and without it the button will submit on click.
- `useTranslation` from `react-i18next` needs to be imported and the two keys (`textShowPassword`, `textHidePassword`) added to `en.json`. The CI merge script will propagate them to the other locales automatically.
- `&__toggle` needs a CSS reset in both `app-ios.less` and `app-material.less`, otherwise the browser's default button border and background will show around the icon:
  ```less
  &__toggle {
      background: none;
      border: none;
      padding: 0;
      cursor: pointer;
      -webkit-appearance: none;
  }
  ```

**Separate issue (iOS only, introduced by this PR):** the new `app-ios.less` block sizes `&__icon` to 24×24 but doesn't size the wrapper. Since this PR is the first to enable the toggle on iOS, the tappable area is whatever the SVG renders — well under the [Apple HIG 44pt target](https://developer.apple.com/design/human-interface-guidelines/buttons). Adding `min-height: 44px; min-width: 44px` to `&__toggle` in `app-ios.less` would cover it (the `&__toggle` reset above is a natural place to add this).

wapec:
@juliusknorr sorry, my mistake in the description. Initially I have pushed the fix for the `type` attribute and alongside changed `span` to `button` with  proper accessibility attributes. but that change required two new translation snippets for `aria-label`. I added them only in English but didn’t know how to proceed with other languages, so the code felt incomplete.

After that I have decided to simplify things by keeping the `type` attribute bug fix + iOS icon support. once these changes were pushed I forgot to update the description.  

regarding testing - yes, I have tested the changes locally with Chrome DevTools. 
the screenshots in the description show two states for both iOS and Android - Password field with IconHide and Verify field with IconShow. I will add this clarification to the description as well.


@moodyjmz I will revert my initial changes for the `button` tag with proper accessibility attributes.  alongside that I will cover missing properties for `&__toggle` in `app-ios.less` and `app-material.less`.

apps/documenteditor/mobile/src/components/PasswordField/PasswordField.jsx
```diff
@@ -18,26 +22,30 @@ const PasswordField = ({ label, placeholder, value, handlerChange, maxLength })
...
-                            </span>
-                        : null}
+                        <input type={isShowPassword ? 'text' : 'password'} placeholder={placeholder} value={password} onChange={handlePasswordChange} className='password-field__input' maxLength={maxLength || null} />
+                        <button
+                            type='button'
+                            onClick={toggleShowPassword}
+                            aria-pressed={isShowPassword}
+                            aria-label={isShowPassword ? t('Settings.textHidePassword') : t('Settings.textShowPassword')}
+                            className='password-field__toggle'
+                        >
```
moodyjmz, line 41: The `<button>` already has an `aria-label`, but `SvgIcon` renders a bare `<svg>` with no `aria-hidden`. Some screen readers will still traverse into it and announce the `<use>` element. Add `aria-hidden` so assistive tech relies solely on the button label.

```suggestion
                            <SvgIcon symbolId={isShowPassword ? IconHide.id : IconShow.id} className='password-field__icon icon icon-svg' aria-hidden={true} />
```

apps/documenteditor/mobile/src/less/app-material.less
```diff
@@ -156,6 +155,19 @@
...
+            padding: 0;
+            margin: 0;
+            cursor: pointer;
+            display: flex;
+            align-items: center;
+            justify-content: center;
+            flex-shrink: 0;
+            width: auto;
+        }
+
```
moodyjmz, line 170: iOS gets a 44×44 px minimum tap target but Material gets none. Android Material guidelines recommend 48 dp; matching iOS at 44 px is a reasonable compromise and keeps both themes consistent.

```suggestion
            width: auto;
            min-width: 44px;
            min-height: 44px;
```

apps/documenteditor/mobile/src/less/app-ios.less
```diff
@@ -101,5 +101,34 @@
...
+            min-height: 24px;
+        }
+
+        &__toggle {
+            background: none;
+            border: none;
+            padding: 0;
+            margin: 0;
+            cursor: pointer;
+            -webkit-appearance: none;
```
moodyjmz, line 123: The unprefixed `appearance` property is missing. `app-material.less` has both on `__input`; keep iOS consistent.

```suggestion
            -webkit-appearance: none;
            appearance: none;
```

moodyjmz (commented):
Three small follow-ups:

PR #66, opened 2026-05-27
feature: add UI support for the checkbox

DmySyz:
add the checkbox directly to the toolbar for now to make access quicker


moodyjmz:
@DmySyz you need to remove the locales from the PR - they get generated and need to be committed elsewhere I think. I haven't had a chance to look to make it better yet I'm afraid.

moodyjmz:
it would be nice to break the commits down logically, no? That way blame shows the why something was done?

moodyjmz:
btw you can run `/code-review`. - this puts Claude into an analysis task mode, vs get shit done mode. Claude is task driven. It will likely surface issues.

PR #69, opened 2026-05-27
fix: correct animation panel icon sizing in presentation editor

### LINKED ISSUE #59: bug: status bar icons don't show in spreadsheet
The current version (test version of may 8) seems to suffer a no-show issue with icons in the status bar, see:

Image

moodyjmz:
SVG icons in the animation effect picker were rendering at 300px wide (browser default) because the .icon:not(svg) CSS rule explicitly excludes SVG elements from the icon size constraint, and the icon uni-scale class has no dimensions of its own.

Replace with explicit width="28" height="28" matching the 28x28 viewBox. This also avoids fill:currentColor overriding the icons' explicit fill colours.

### DIFF
--- a/apps/presentationeditor/main/app/view/Animation.js
+++ b/apps/presentationeditor/main/app/view/Animation.js
@@ -201,7 +201,7 @@ define([
                     autoWidth:       true,
                     itemTemplate: _.template([
                         '<div  class = "btn_item x-huge" id = "<%= id %>" style = "width: ' + itemWidth + 'px;height: ' + itemHeight + 'px;">',
-                            '<svg class="icon uni-scale"><use href="#<%= iconCls %>"></use></svg>',
+                            '<svg width="28" height="28"><use href="#<%= iconCls %>"></use></svg>',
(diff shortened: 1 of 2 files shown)


chrip (approved):
Thanks for the fix—this is a clean, targeted change and the PR description clearly explains the root cause.

Minor side note: the branch name `fix/presentation-click-offset` doesn't quite match the content, but that's trivial.

LGTM. The explicit `width`/`height` on the `<svg>` elements is the right approach and avoids the CSS selector gap (`:not(svg)`) that was causing the 300px overflow. Approved.

chrip (approved):
Thanks for the fix—this is a clean, targeted change and the PR description clearly explains the root cause.

Minor side note: the branch name `fix/presentation-click-offset` doesn't quite match the content, but that's trivial.

LGTM. The explicit `width`/`height` on the `<svg>` elements is the right approach and avoids the CSS selector gap (`:not(svg)`) that was causing the 300px overflow. Approved.

PR #70, opened 2026-05-29
feat: add edit image from storage on mobile editors

n-goncalves:
This PR allow to add and replace existing images using the storage server on mobile for documenteditor, presentationeditor and spreadsheeteditor.  

image

Edit: force-pushed to close the modal when button is clicked on as that's what happens when choosing a local file.


apps/documenteditor/mobile/src/controller/Main.jsx
```diff
@@ -589,7 +589,12 @@ class MainController extends Component {
             data._urls = arr;
         }
 
-        this.insertImageFromStorage(data);
+        if(data.c === 'add') {
+            this.insertImageFromStorage(data);
+        }
+        else if(data.c === 'change') {
+            this.replaceImageFromStorage(data);
+        }
```
moodyjmz: Before this change, `insertImageFromStorage(data)` ran unconditionally and its guard treats a missing `c` as add (`!data.c || data.c === 'add'`), same as the desktop `Toolbar.js`. With the strict `=== 'add'` here, an integrator calling `insertImage({images: [...]})` without `c` used to insert and now silently does nothing.

```suggestion
        if (!data.c || data.c === 'add') {
            this.insertImageFromStorage(data);
        }
        else if (data.c === 'change') {
            this.replaceImageFromStorage(data);
        }
```

moodyjmz (commented):
Thanks @n-goncalves — looks good overall, and it matches how the desktop editors do it (DE/PE `ImgApply`, SSE `asc_setGraphicObjectProps`; `put_ImageUrl` is a valid alias of `asc_putImageUrl` in the cell SDK, so that's fine too). Nice touch putting the icon in `common/`.

One behavioural regression flagged inline — the missing-`c` fallback to "add" got dropped. It applies to all three editors; suggestions attached so it's a one-click fix.

Couple of non-blocking nits: `changeCanRequestInsertImage` is registered in `makeObservable` in the three stores but never called (the value is assigned directly in `setConfigOptions`), and indentation is off in the new JSX blocks in DE/PE `view/add/AddImage.jsx`.

apps/presentationeditor/mobile/src/controller/Main.jsx
```diff
@@ -647,7 +655,12 @@ class MainController extends Component {
             data._urls = arr;
         }
 
-        this.insertImageFromStorage(data);
+        if(data.c === 'add') {
+            this.insertImageFromStorage(data);
+        }
+        else if(data.c === 'change') {
+            this.replaceImageFromStorage(data);
+        }
```
moodyjmz: Same as in the document editor — missing `c` should fall through to add.

```suggestion
        if (!data.c || data.c === 'add') {
            this.insertImageFromStorage(data);
        }
        else if (data.c === 'change') {
            this.replaceImageFromStorage(data);
        }
```

apps/spreadsheeteditor/mobile/src/controller/Main.jsx
```diff
@@ -795,7 +803,12 @@ class MainController extends Component {
             data._urls = arr;
         }
 
-        this.insertImageFromStorage(data);
+        if(data.c === 'add') {
+            this.insertImageFromStorage(data);
+        }
+        else if(data.c === 'change') {
+            this.replaceImageFromStorage(data);
+        }
```
moodyjmz: Same as in the document editor — missing `c` should fall through to add.

```suggestion
        if (!data.c || data.c === 'add') {
            this.insertImageFromStorage(data);
        }
        else if (data.c === 'change') {
            this.replaceImageFromStorage(data);
        }
```

apps/documenteditor/mobile/src/controller/Main.jsx
```diff
@@ -1045,6 +1050,14 @@ class MainController extends Component {
         }
     }
 
+    replaceImageFromStorage (data) {
+        if (data && data._urls && (!data.c || data.c === 'change') && data._urls.length > 0) {
```
moodyjmz: With the dispatch above fixed, the `!data.c` here becomes unreachable — and on its own it documents the wrong intent: a missing command should never mean replace (anyone calling this method directly would replace an image on a command that never asked for it).

```suggestion
        if (data && data._urls && data.c === 'change' && data._urls.length > 0) {
```

apps/presentationeditor/mobile/src/controller/Main.jsx
```diff
@@ -550,6 +550,14 @@ class MainController extends Component {
         }
     }
 
+    replaceImageFromStorage (data) {
+        if (data && data._urls && (!data.c || data.c === 'change') && data._urls.length > 0) {
```
moodyjmz: Same here — drop the `!data.c`; a missing command shouldn't trigger a replace.

```suggestion
        if (data && data._urls && data.c === 'change' && data._urls.length > 0) {
```

apps/spreadsheeteditor/mobile/src/controller/Main.jsx
```diff
@@ -613,6 +613,14 @@ class MainController extends Component {
         }
     }
 
+    replaceImageFromStorage (data) {
+        if (data && data._urls && (!data.c || data.c === 'change') && data._urls.length > 0) {
```
moodyjmz: Same here — drop the `!data.c`; a missing command shouldn't trigger a replace.

```suggestion
        if (data && data._urls && data.c === 'change' && data._urls.length > 0) {
```

moodyjmz (commented):
Left some comments

n-goncalves:
Thanks for the catch on the empty `c` arguments, it wasn't mentioned on the API documentation. 

For the minor nitpicks, I have fixed the indentation in the new JSX blocks in DE/PE view/add/AddImage.jsx.
I have also removed the `changeCanRequestInsertImage` in appOptions. 

I wrote the first iteration of this years ago, I assume I wanted to do something with it at first but never did.

moodyjmz (commented):
Thanks for addressing all the previous round's suggestions — the dispatch fallback and the guard corrections are both clean.

Two things still to tidy:

**1. PE `view/add/AddImage.jsx` — `</List>` indent**

The closing tag landed at 8-space indent while `<List>` and everything inside it sits at 12. DE got this right; PE just needs the closer brought in one level.

**2. SSE `storeAppOptions.js` — `canRequestInsertImage` in `makeObservable`**

Worth a quick check: the SSE store uses an explicit `makeObservable` call that enumerates each observable field. The diff doesn't show `canRequestInsertImage` being added to that block — only the class field and the `setConfigOptions` assignment. If it's actually absent, MobX won't track it for reactivity (plain JS property, observer components won't re-render on change). Functionally fine for a config-once value, but worth confirming the intent is deliberate — or adding `canRequestInsertImage: observable` to match the other config flags in that block.

Minor nit: there's a stray double blank line after `onInsertByStorage` in the PE and SSE add controllers — single blank is enough.

n-goncalves:
> 1. PE view/add/AddImage.jsx — </List> indent

Resolved.

> 2. SSE storeAppOptions.js — canRequestInsertImage in makeObservable

The `canRequestInsertImage` value in appOptions comes from the fact that the `onRequestInsertImage` event exists in the editor when creating the DocEditor in [api.js#396](https://github.com/Euro-Office/web-apps/blob/main/apps/api/documents/api.js#L396)
As that is supposed to be set only once, it should be fine leaving it as it is. 

But if I look at similar variables like for instance `canBrandingExt` in the same file, that one is also set once, but is in the `makeObservable` so using the same style also makes sense. I've let it as is for now but am happy to change it. 

You mention only the SSE `storeAppOptions.js` but that's also the case in DE and PE. 

> Minor nit: there's a stray double blank line after onInsertByStorage in the PE and SSE add controllers — single blank is enough.

Resolved for PE add controller. I have checked the SSE add controller but there is a single blank line there.

moodyjmz:
Looking good, one tiny, tiny thing

canRequestInsertImage is set once, however I think it should be observable like the others, per editor.

`[editor]/src/store/appOptions.js`

```js
makeObservable(this, {
    ....
    canRequestInsertImage: observable,   // <<< add this
});
```

Once that's done, I'll approve and merge - if you agree with those changes :)

moodyjmz (approved):

moodyjmz:
Looks good, thanks for making those changes, sorry for the delay getting round to reviewing it @n-goncalves

PR #91, opened 2026-06-12
ci: e2e job overlaying fresh web-apps build on nightly image

juliusknorr:
Adds an `e2e` workflow that builds web-apps, overlays the fresh build onto `ghcr.io/euro-office/documentserver:nightly`, and runs the DocumentServer Playwright e2e suite against it. Full-stack feedback on a web-apps change without rebuilding core/server/sdkjs.

## How it works
- Pulls the nightly image and reads the `PRODUCT_VERSION` it was built with (from the bundled `api.js`).
- Builds web-apps (`THEME=euro-office grunt`, mirroring the bake) with that `PRODUCT_VERSION`.
- Overlays the fresh `web-apps` onto the image (install path discovered via the `sdkjs` dir, not hardcoded) and runs the e2e suite from `Euro-Office/DocumentServer` (matching branch, fallback `main`).

All involved repos and the documentserver package are public, so the image pulls anonymously (no PAT).

## Why the version is injected
The editor reports a build version to docservice, which rejects the document when it differs from its own (exact-match gate) — the editor loads but the document never opens. A fresh build defaults to the package.json version (`1.0.1`), so it must be built with the same version as the nightly image. The version is read from the image so it tracks bumps automatically.

Verified green against the current nightly.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,120 @@
+name: E2E (overlay on nightly)
+
+# Builds web-apps, overlays the fresh build onto the published
+# ghcr.io/euro-office/documentserver:nightly image, and runs the DocumentServer
+# Playwright e2e suite against it. Fast feedback on a web-apps change against
+# the full stack without rebuilding core/server/sdkjs.
+on:
+  push:
+    branches:
+      - fork
+      - develop
+      - 'release/**'
+      - 'hotfix/**'
+  pull_request:
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  NIGHTLY_IMAGE: ghcr.io/euro-office/documentserver:nightly
+  OVERLAY_IMAGE: euro-office/documentserver:e2e-web-apps
+
+jobs:
+  e2e:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Check out web-apps
+        uses: actions/checkout@v4
+        with:
+          path: web-apps
+
+      # The e2e suite lives in the DocumentServer repo. Prefer a same-named
+      # branch (so tests can be co-developed with the change), else main.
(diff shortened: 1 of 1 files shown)


chrip (approved):
Thanks a lot for this, @juliusknorr! This is a really clever approach — overlaying fresh web-apps on the nightly image to get full-stack e2e feedback without rebuilding the entire DocumentServer is exactly the kind of thing we need. The version injection via `api.js` is especially well thought out.

A few things to consider:

1. **`push` trigger doesn't include `main`** — Lines 9-13 list `fork`, `develop`, `release/**`, `hotfix/**`. PRs to `main` are covered, but a direct push to `main` (revert, force push, hotfix merge) would skip e2e entirely. Worth adding if direct-to-main pushes are possible.

2. **`find` patterns could be more specific** — Line 85's `find /var/www -maxdepth 8 -path '*api/documents/api.js'` and line 109's `find ... -name sdkjs -path '*documentserver*'` rely on `head -n1` when multiple matches could exist. This works for current image layouts but is fragile. Consider anchoring closer to the expected path structure if DocumentServer has a stable layout.

3. **Missing `timeout-minutes`** — The job has no timeout. If Playwright tests hang (stalled browser, Docker issues), the job can run up to 6 hours. Adding `timeout-minutes: 30` (or whatever feels right) would be a safety net.

4. **No npm/Playwright caching** — `npm install`, `npm ci`, and `npx playwright install chromium` run fresh every time. With the nightly image pull (~1-2GB) and browser download (~300MB), caching via `actions/cache` would significantly cut run time and cost.

5. **Mutable `:nightly` tag** — The `:nightly` image changes independently, so results aren't reproducible. If you ever need to bisect a failure, knowing exactly which image was tested would help. Logging the image digest after `docker pull` would capture this.

Overall this is a great contribution — the overlay strategy, version extraction, and DocumentServer branch matching logic are all solid. The points above are improvements, not blockers.

Assisted-by: OpenCode:Qwen3.6-27B

moodyjmz (approved):
once we move to web pack, this needs switching to that

Issue #88, opened 2026-06-11
feat Mobile: Implement context menu depending on selection


n-goncalves:
For instance, trying to select text, in this case, you'd expect being able copy/cut/paste the selection and maybe some other action depending on which element is selected. 

Currently you have none of that, you might get one or two context action that are added from somewhere else like in the case linked below. 

<img width="362" height="644" alt="Image" src="https://github.com/user-attachments/assets/3afe4998-3669-47cc-9b96-fc5c4699bd57" />

That's because the `patch.jsx` is missing the `ContextMenu` [implementation](https://github.com/Euro-Office/web-apps/blob/main/apps/documenteditor/mobile/src/lib/patch.jsx#L17C4-L17C34). I haven't checked the other editors but I assume it is the same. 

I have a working implementation that was based on the old mobile editor before they removed it. The quality is not good enough to take it as is, but it will serve as a good starting point to see what would be the specs of this feature if that's something you would be interested in. I can create a draft PR in that case.

moodyjmz:
@n-goncalves interesting, from the comment in the code it wasn't pulled over to "retain parity". I need to revisit to see what that means.

n-goncalves:
I can't answer the why it was commented out but I have removed the comments to see what's the current state of `ContextMenu` in `editor.jsx`, focusing on DE for now as PE and SSE both seem to have more mature implementation. 

- Size of context bar popover is fixed regardless of the number of actions. 

<img width="437" height="212" alt="Image" src="https://github.com/user-attachments/assets/bd0a8a02-945e-4331-b3d9-71a7f4e24725" />

- `handleMenuItemClick` always return false, triggering the fallback path. Other editors do have working implementations of that method so taking the one from PE should get us most if not all the way there. 

- `mapMenuItems` these are the contextual actions currently implemented
  - Copy/Cut/Paste 
  - Add or View Comment
  - Open and Edit Link

If you look at old [DE code](https://github.com/Euro-Office/web-apps/commit/b0c9a1c5d28371d78ae54854f74f9bdaa8ee9efa) before they pulled mobile or at PE/SSE they are more actions possible. 

Would the team accept PR(s) that fixes all the above ? 

Additionally there is a issue with contextual actions in all editors. The actions to show are returned by `mapMenuItems` and additionnal actions are stored in `controller.extraItems`. Currently the split is done by the device type and the number of actions. If you happen to have an unfortunate combo of device width, number of actions and text length, the popout is too big and goes past the screen.

[...]

PR #98, opened 2026-06-13
DE/PE mobile: Fix getEditCommentControllers using already added controllers

n-goncalves:
While working on a draft PR for #88, I found out that the `getEditCommentControllers` in editor.jsx were implemented incorrectly for documenteditor and presentationeditor. (spreadsheeteditor doesn't have that method and is therefore fine)

It is adding controllers that are added in Main.jsx ([documenteditor](https://github.com/Euro-Office/web-apps/blob/53dd1a19a7a9a7dc670545e6a04443b7b8a31ea1/apps/documenteditor/mobile/src/controller/Main.jsx#L1682-L1684) and [presentationeditor](https://github.com/Euro-Office/web-apps/blob/53dd1a19a7a9a7dc670545e6a04443b7b8a31ea1/apps/presentationeditor/mobile/src/controller/Main.jsx#L1213-L1215)), so it wasn't doing anything. I added the expected controllers to allow adding and editing comments. 

Right now it doesn't do anything as it needs the context menu actions to be shown and handled in order to open the comments but here's the diff from before ONLYOFFICE removed their mobile editor to show that's what's supposed to be there:
* documenteditor: https://github.com/Euro-Office/web-apps/commit/45a9058da461a9da67ccb389faad08bce8cbda55#diff-e02881fa2158c2909c0fb78180c1e55f2cc41db27579bfcc1e7844bc4e0f03bdL785-R775
* presentation: https://github.com/Euro-Office/web-apps/commit/45a9058da461a9da67ccb389faad08bce8cbda55#diff-1d89c904c6cd5147fb6942c09c4d12a2f3a3200a61b3e29b1a42a1263e57848bL307-R303

### DIFF
--- a/apps/documenteditor/mobile/src/lib/editor.jsx
+++ b/apps/documenteditor/mobile/src/lib/editor.jsx
@@ -1,6 +1,6 @@
 import React, { Fragment } from 'react';
 import { Link } from 'framework7-react';
-import { CommentsController, ViewCommentsController } from '../../../../common/mobile/lib/controller/collaboration/Comments';
+import { AddCommentController, EditCommentController } from '../../../../common/mobile/lib/controller/collaboration/Comments';
 import {
     PlatformIcon,
     buildFocusObjectGetters,
@@ -172,12 +172,12 @@ export const updateChartStyles = (storeChartSettings, storeFocusObjects) => {
 
 /**
  * Renders comment controller components for editing mode
- * @returns {JSX.Element} Fragment containing CommentsController and ViewCommentsController
+ * @returns {JSX.Element} Fragment containing AddCommentController and EditCommentController
  */
 export const getEditCommentControllers = () => (
     <Fragment>
-        <CommentsController />
-        <ViewCommentsController />
+        <AddCommentController />
+        <EditCommentController />
     </Fragment>
 );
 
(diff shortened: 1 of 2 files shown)


moodyjmz (approved):
> _Review authored by **Claude (Opus 4.8)**, reviewed and approved by @moodyjmz. Verified against `origin/main`._

Thanks @n-goncalves — nice catch, and good to have you back. 👍

This is a clean fix for a real bug. `getEditCommentControllers()` was returning `CommentsController` + `ViewCommentsController` — the same two controllers already mounted directly in `Main.jsx` (DE `1682`/`1684`, PE `1213`/`1215`). So in edit mode those two were mounted twice, while `AddCommentController` / `EditCommentController` — the ones a function named *EditComment* should actually provide — were never mounted at all in DE/PE mobile. The swap fixes both: no more duplicate mounts, and add/edit-comment now has its controllers, matching the pattern the spreadsheet editor already uses.

Verified:
- `AddCommentController` / `EditCommentController` are real exports of `common/mobile/lib/controller/collaboration/Comments.jsx`.
- No dangling references — `CommentsController`/`ViewCommentsController` are no longer used in either `editor.jsx`, so dropping the import is clean.
- Symmetric across DE and PE, JSDoc updated to match.

Approving and merging. Thanks for the contribution!

moodyjmz:
@n-goncalves Nice clean work, thanks again!

Issue #108, opened 2026-06-15
docs: add .claude/ folder to repo for topology and session context


moodyjmz:
## Background

During the webpack migration, a `.claude/` folder was used locally to store:
- `migration-topology.md` — every file changed, commit refs, known gaps
- `findings/*.md` — individual technical findings (non-obvious gotchas that took significant debugging)
- Other session context docs (docker topology, icon migration notes, etc.)

These files are currently `.gitignore`d (local only). This means:
- A new contributor or session has no access to the hard-won debugging findings
- If the local copy is lost, the institutional knowledge is gone
- The findings are referenced from `CLAUDE.md` but can't be followed

## Proposal

Commit `.claude/` to the repo (remove from `.gitignore`). This gives:
- Persistent, reviewable technical findings alongside the code they describe
- AI session context that survives machine changes and onboarding new contributors
- A natural home for topology docs (docker setup, multi-repo relationships) that are too detailed for README but too important to leave local

The `.claude/findings/` files in particular are high-value — they document non-obvious constraints that would take hours to re-discover (e.g. `mangle:false` load-bearing behaviour, locale.js AMD parser crash, inline-svgs ordering invariant).

## Scope

- Remove `.claude/` from `.gitignore`
- Review and commit existing findings files (redact anything session-specific)
- Update `CLAUDE.md` to reference committed paths

Linked from: webpack migration PR (see PR that adds `docs/webpack-migration.md`)

moodyjmz:
Related PR that adds docs/webpack-migration.md (the committed replacement for local-only findings): Euro-Office/web-apps#109

Issue #106, opened 2026-06-15
build: replace grunt-inline SVG/script embedding with cached external assets


moodyjmz:
## Problem

The build bakes SVG sprite files and small utility scripts directly into each editor's `index.html` at build time (via `inline-svgs.js`, which replicates grunt-inline behaviour). This has two downsides:

1. **Duplicate copyright headers** — each inlined file carries its own AGPL copyright block; the deployed HTML contains it multiple times in the page source.
2. **No independent caching** — SVG sprites (hundreds of KB of paths) and JS snippets are re-sent with every page load and cannot be cached separately from the app HTML. Any change to a sprite invalidates the whole page.

## Current flow

`index.html.deploy` uses grunt-inline conventions:
- `<inline src="...svg" />` → SVG content embedded verbatim
- `<script src="...js?__inline=true">` → JS content embedded verbatim

`inline-svgs.js` (our grunt-inline replacement) performs these substitutions at build time.

## Suggested approaches

- Serve SVG sprites as versioned static files with long-lived cache headers; reference via `<img src="sprites.svg?v=BUILD_NUMBER">` or a `<link rel="preload">`.
- Move the inlined utility scripts (`desktopinit.js`, `themeinit.js`, `htmlutils.js`, `docserviceworker.js`) to versioned external files with cache-busting query strings.

## Scope

Out of scope for the current webpack migration (`build/webpack-migration`). Track here for a follow-up build hygiene pass.

## Related files

- `build/scripts/inline-svgs.js`
- `apps/*/main/index.html.deploy` (all editors)
- `apps/common/main/index.html.deploy`

Issue #89, opened 2026-06-12
Editor loader still shows ONLYOFFICE branding


michidold:
If a document is opened it still shows the OnlyOffice loading logo. If I inject a custom one to override it (editorConfig.customization.loaderLogo/loaderName), I get a white-label / "paid feature" license dialog. We're using @onlyoffice/document-editor-react btw.

Seen on Version: 9.3.1  (ghcr.io/euro-office/documentserver)

<img width="468" height="551" alt="Image" src="https://github.com/user-attachments/assets/4bea3f9d-4264-4afa-96be-cf68fabcf4a4" />

Generally there are several places where it still says OnlyOffice, which is hard to explain to customers — e.g. the document editor index.html still has <title>ONLYOFFICE Documents</title>.

moodyjmz:
Thanks for this, something to investigate, I am curious about the react bit, as this suggests this is something none standard

juliusknorr:
Is the loading screen maybe coming from the react library? I have not seen it in our tests anywhere.

michidold:
> Is the loading screen maybe coming from the react library? I have not seen it in our tests anywhere.

No, double checked that -> think its from https://github.com/Euro-Office/web-apps/blob/main/apps/documenteditor/main/index_loader.html

PR #109, opened 2026-06-15
build: replace Grunt+r.js with webpack 5 — Phase E complete

### LINKED ISSUE #107: build: evaluate esbuild as TerserPlugin replacement for mobile (framework7-react) builds
## Background

Mobile builds (`vendor/framework7-react`) use webpack 5 + Babel + TerserPlugin. The desktop editors cannot use esbuild for minification because `mangle: false` is load-bearing (117 files use `var Common = Common || {}` namespace guards that break under name mangling). Mobile has no such constraint — it is a self-contained React/webpack bundle with no shared globals.

## Why esbuild is feasible for mobile

- **ES target is modern**: `browserslist` targets Chrome 49+, iOS 11+, Android 7+; babel config uses `> 0.25%, not dead` which resolves to Chrome ~110+, Firefox ~100+, Safari ~15+ — firmly ES6+.
- **No `mangle: false` requirement**: mobile bundle has no AMD/global namespace guards; full name mangling is safe.

## Tested ✓

Swapped `TerserPlugin` for `EsbuildPlugin` from `esbuild-loader` 4.5.0 in `vendor/framework7-react/build/webpack.config.js`:

```js
import { EsbuildPlugin } from "esbuild-loader";

minimizer: [
  new EsbuildPlugin({
    target: 'es2019',
    drop: env === 'production' ? ['console'] : [],
  }),
  new CssMinimizerPlugin({ ... }),
]
```

Babel stays for transpilation (`@babel/plugin-proposal-decorators` legacy mode — esbuild does not support legacy decorators).

**Results (in `eo` container):**

| | TerserPlugin | EsbuildPlugin |
|--|--|--|
| Single editor | ~90s | 25s |
| All 4 parallel | ~90s | **44s** |

~2× faster on wall clock for the parallel build. No new warnings — the 3 webpack size hints are pre-existing.

## Blocker

`vendor/framework7-react/` is gitignored in this repo and managed separately. The change cannot land here. This work needs to be applied wherever that directory is canonically tracked.

## Files to change (in the framework7-react repo)

- `build/webpack.config.js` — swap TerserPlugin import + minimizer entry as above
- `package.json` — add `esbuild-loader` to devDependencies
- `npm-shrinkwrap.json` — updated lockfile

moodyjmz:
## TL;DR

Grunt is gone. All build tasks now run through webpack 5 + Node.js scripts.

**CI build time: ~7 minutes → ~2 minutes** (cold runner, no cache).  
**Local build time (eo container, SKIP_MOBILE=1)**: ~101s.  
**Mobile build time per editor: ~146–179s → ~101–132s** (EsbuildPlugin, see below).  
All 6 editors (documenteditor, spreadsheeteditor, presentationeditor, visioeditor, pdfeditor, forms) runtime-validated. CI green.

Developed with [Claude Code](https://claude.ai/code) — AI-assisted analysis, debugging, and implementation.

---

## Testing

Pull https://github.com/Euro-Office/DocumentServer/pull/187

pull this PR, run as normal `make local`, then in docker (once) `make sdkjs` and `make web-apps-dev`, see if errors come up. Test Euro Office in browser. Look for console errors or anything else.

## Build system improvements (2026-06-17)

Three follow-up commits on top of the Phase E work, addressing mobile build performance and pipeline robustness.

<details>
<summary>⚡ EsbuildPlugin replaces TerserPlugin + CssMinimizerPlugin (closes #107)</summary>

**Why:** Mobile webpack builds were taking 146–179s per editor with TerserPlugin doing JS minification and CssMinimizerPlugin doing CSS minification in two separate passes. esbuild's minifier handles both in a single pass and is significantly faster.

**What changed:** `vendor/framework7-react/build/webpack.config.js` — the two-entry `optimization.minimizer` array and the duplicate `CssMinimizerPlugin` in the production plugins array are replaced with a single `new EsbuildPlugin({ target: 'es2015', css: true })`. `terser-webpack-plugin` and `css-minimizer-webpack-plugin` removed from `package.json`; `esbuild-loader` added.

**Validation — two full pipeline runs in `eo` container:**

| Task | Before | After run 1 | After run 2 |
|------|--------|-------------|-------------|
| mobile:word | 146–179s | 130.4s | 132.9s |
| mobile:cell | 146–179s | 127.2s | 130.8s |
| mobile:slide | 146–179s | 122.5s | 126.1s |
| mobile:visio | 146–179s | 101.4s | 116.2s |
| Wall clock | — | 176.9s | 164.0s |

All 21 pipeline tasks green both runs. The remaining mobile build time (~100–130s) is babel-loader + bundling — the minification step is no longer the bottleneck.

**Note:** Further improvement is possible by replacing babel-loader with esbuild-loader for JS transforms (~30–50s additional saving). Deferred — the mobile code uses MobX `@inject` legacy decorators which require `experimentalDecorators: true` and warrant a separate tested commit.

</details>

<details>
<summary>✨ Lazy task spawn + word-wrapped phase task list</summary>

**Why:** `task()` previously spawned child processes eagerly. Phase 1 builds `phase1Tasks` (deploy ×5 + webpack ×6) eagerly, then awaits `mobile:install` (~20s), then calls `phase()`. This meant the phase banner and task list were printed *after* the fast deploy tasks had already completed and printed their ✓ lines — output order was banner-after-results, which was confusing.

**What changed:** `task()` now returns a lazy spec (`{ label, cmd, args, opts }`). `runTask(spec)` spawns and returns `{ promise, kill, label }`. `phase()` prints the banner and task list first, then calls `runTask()` on each spec before awaiting. Output order is now always correct: header → task completion lines.

The task list in the banner is also reformatted: previously one long comma-separated line (`sprites, deploy-common, deploy-html, ...`); now word-wrapped at 80 columns with ` · ` separators.

</details>

<details>
<summary>🐛 Flush stderr of killed tasks on sibling failure</summary>

**Why:** When a phase task fails, `phase()` sends SIGTERM to all sibling processes. Killed tasks resolved with `code: -1` but their buffered stderr was discarded — if the killed task had begun emitting a useful error before being killed, that output was lost.

**What changed:** One line in `runTask()`: `stderrBuf` is now flushed in the `signal` branch of `child.on('exit')`, matching the existing flush in the non-zero exit branch.

</details>

---

## CI fix — e2e workflow switched from grunt to build-pipeline.js (2026-06-18)

PR #91 (merged to `main`) added `.github/workflows/e2e.yml` with a build step that calls `grunt`. This PR removes grunt. The two collide: `e2e.yml` would fail with `grunt: command not found` the moment #109 landed.

**Fix:** merged `origin/main` into the branch (clean, no conflicts — `e2e.yml` was a new file) then updated the build step:

- **Dropped:** `npm install -g grunt-cli` and `grunt`
- **Added:** `node scripts/build-pipeline.js` and `SKIP_MOBILE: '1'`
- **Unchanged:** `BUILD_ROOT`, `THEME`, `PRODUCT_VERSION` env vars; the version-read step; the overlay build step; all triggers and Playwright run steps

`SKIP_MOBILE: '1'` is correct here — the e2e suite runs Playwright on Chromium desktop only; the framework7 mobile editors are not exercised, so skipping them saves ~130s of CI time per run.

**Bonus:** `build-pipeline.js` runs the Tier-1 gates (verify-replacements preflight + verify-bundles/verify-deploy final gate). The e2e job now fails fast on a broken shim, unreplaced token, or missing output artifact — before Playwright spins up the nightly image. Cheaper failure, earlier signal.

---

## What changed

Six webpack configs replace r.js AMD bundling. Six Node.js scripts replace all grunt asset tasks. A parallel orchestrator (`build/scripts/build-pipeline.js`) replaces the sequential Makefile chain. Grunt, Gruntfile.js, and all grunt test fixtures are removed.

<details>
<summary>Pipeline overview</summary>

**Phase 1 (all parallel):** sprites, deploy-common, deploy-html, deploy-reporter, deploy-theme-images, deploy-embed, webpack ×6, mobile ×4

**Phase 2 (after Phase 1):** deploy-resources, deploy-theme-img, inline-svgs

Local builds get webpack filesystem cache: 19s cold → 0.6s warm rebuild per editor (full pipeline wall-clock is dominated by other Phase 1 tasks).

</details>

<details>
<summary>Theme LESS overrides</summary>

grunt's `appendThemeFiles()` injected `theme/euro-office/assets/less/theme.less` as a top-level LESS entry. webpack equivalent:

1. `theme.config.mjs` writes a one-line redirector stub at `apps/common/main/resources/less/_theme-main.less`
2. Each editor's `app.less` imports the stub at the end

The stub is a redirector (not a copy) so `theme.less`'s own relative `@import "overrides.less"` resolves from the theme directory. Adding a new theme override is theme-only — no build changes required. See `docs/theme-less-overrides.md`.

CSS outputs to `apps/<editor>/main/app.css` (root of `main/`), matching grunt's baseline so all `url()` references in LESS resolve at the correct depth.

</details>

<details>
<summary>Gotchas worth knowing</summary>

These surfaced during migration and are documented in `docs/webpack-migration.md`:

- **`mangle: false` is load-bearing** — 117 files use `var Common = Common || {}` as a namespace guard that breaks under name mangling in webpack's module scope. Not negotiable without a larger refactor.
- **`PRODUCT_VERSION` must be ≥ 6** — EuroOffice rejects editors below version 6. The `common.json` fallback is `4.3.0`. Pipeline fails fast if it's missing.
- **`locale.js` crashes webpack's AMD parser** — fixed with `string-replace-loader`. Do not use `noParse`.
- **`deploy-html` + `inline-svgs` are a unit** — running deploy-html alone leaves broken `?__inline=true` tags and produces blank editors.
- **SVG sprites are baked into HTML at build time** — no XHR requests for icons at runtime. `index.html` (un-built source) still uses SVGInjector but that file is never deployed.
- **Service worker caches aggressively** — always test in incognito.

</details>

<details>
<summary>Known pre-existing issues (not caused by this migration)</summary>

| Issue | Notes |
|-------|-------|
| `warnings_s.svg` 404 | CSS `url()` path depth — pre-existing |
| Transitions panel icons blank | CSS classes not updated for SVG migration |
| `FormsTab.getView()` throws on plain PDF | Upstream OnlyOffice bug, SDK catches it |
| Large locale diffs | CI translation-merge step — expected noise, not regressions |

</details>

## Tracked follow-ups

- #106 — Replace grunt-inline SVG/script embedding with cached external assets
- ~~#107 — Evaluate esbuild as TerserPlugin replacement for mobile builds~~ ✅ resolved in this PR
- #108 — Add `.claude/` findings folder to repo
- ~~Update `e2e.yml` build step from grunt to `build-pipeline.js`~~ ✅ resolved in this PR (2026-06-18)

## DocumentServer changes

Corresponding Makefile update (removes grunt from `make web-apps-dev`, calls `build-pipeline.js`): see Euro-Office/DocumentServer PR [linked below].


moodyjmz:
DocumentServer Makefile companion PR: Euro-Office/DocumentServer#187

moodyjmz:
Just to add the plan:

---
title: Build System Migration Plan — Option D (webpack 5 consolidation)
date: 2026-06-12
status: in progress
---

# Build System Migration Plan

Incremental consolidation of the desktop editor front-end onto webpack 5.
Full context: see `build-system-report.md`.

---

## Full goal

**Remove grunt entirely from the build pipeline.**

This is broader than replacing r.js. Grunt currently owns:

| Grunt task | Replacement | Status |
|---|---|---|
| r.js bundling — desktop `app.js` | our webpack configs | ✓ done (doc/spreadsheet/presentation/visio) |
| SVG sprite *generation* (`sprites.sh` / `prebuild-svg-sprites`) | `sprites.sh` already standalone — CI calls it before grunt. `prebuild-svg-sprites` runs a different task (`all-svg` — 5 named targets) than `sprites.sh` default (`svg_sprite` — all targets); not simply redundant. The `icons.svg` output files are git-tracked (committed); grunt-inline reads whichever version is on disk. When grunt is retired, `sprites.sh` stays in CI; `prebuild-svg-sprites` is removed — but confirm `all-svg` vs `default` produce equivalent output first. | ✓ generation already standalone |
| SVG sprite *injection* (`grunt-inline`) — embeds `icons.svg` content into HTML (`<inline src="...icons.svg"/>` → inline SVG markup in `index.html`) | small Node post-build script: read `icons.svg` files (git-tracked, regenerated by `sprites.sh`) and replace `<inline src="..."/>` tags in built HTML. No webpack plugin needed. | pending — blocks grunt retirement |
| `deploy-theme-images` — copies themed images into `apps/<editor>/mobile/resources/img` | webpack `CopyPlugin` rule | pending — desktop + mobile both depend on this |
| `deploy-app-embed` — embed editor builds (terser + less + copy + inline) for doc/spreadsheet/presentation/visio. **pdfeditor has no embed.** | direct npm script or webpack config | pending |
| `deploy-app-forms` (in `appforms.js`) — forms build, pulled in via documenteditor's deploy list | direct npm script or webpack config | pending |
| `exec:webpack_app_build` + `exec:webpack_install` — mobile Framework7 | call `npm run deploy-<editor>` (or `node build/build.js`) directly from CI / Makefile; preserve `npm install` in `vendor/framework7-react` first | pending |
| `grunt-text-replace` — brand tokens | `string-replace-loader` | ✓ done |
| Translation merge | `python3 merge_and_check.py` (already standalone) | ✓ done |
| `grunt-contrib-copy` / `grunt-svgmin` | webpack `CopyPlugin` / svgo CLI | pending (absorbed by other items above) |

**Mobile webpack is out of scope for modification.** The Framework7 mobile build already uses webpack internally (`vendor/framework7-react/build/webpack.config.js`). Grunt currently orchestrates it via `exec:webpack_app_build` which runs `node ./build/build.js` — a wrapper that does `rimraf ./www/` + `webpack(config, cb)` with a `TARGET_EDITOR` env mapping (`word/cell/slide/visio`). The `exec:webpack_install` step runs `npm install` in the vendor dir first. **Calling "mobile webpack directly" means running `node build/build.js` (not invoking the config directly) and preserving the `npm install` and `TARGET_EDITOR` mapping.** The mobile build internals are not touched.

---

## CI vs Makefile — separate orchestrators

**CI** (`.github/workflows/build.yml`) does **not** use the Makefile. It calls each step directly:
```
sprites.sh → npm install → python3 merge_and_check.py → grunt → webpack
```
When grunt is retired from CI, its step is removed and mobile webpack is added as a direct call. No Makefile involved.

**Makefile** (`DocumentServer/develop/setup/Makefile`, a separate repo) is used inside the `eo` container for local dev only. `make web-apps-dev` triggers the full build. Both `web-apps` and `web-apps-dev` targets now call webpack after grunt (updated 2026-06-14).

These two orchestrators must be kept in sync manually — there is no single source of truth for the build sequence. This is a known maintenance risk.

**Makefile is in the DocumentServer repo**, not web-apps. Changes to it require a separate branch/PR in that repo.

---

## Guiding principle

Every step ends with: **build → structural diff → e2e green**.
Nothing merges without a passing gate.

---

## Step 1 — Establish the validation baseline (NOW)

### Findings: there is no e2e test suite

The CI workflow (`.github/workflows/Build.yml`) does:
1. Check out `web-apps` + `sdkjs`
2. Run `sprites.sh`
3. `npm install`
4. Run translation key merge/check
5. `grunt`

That is it. **CI validates that Grunt runs without error, not that the output works.**

`build/package.json` has `mocha` as a devDep but there are no test files in the project source — only in `node_modules`. There is no e2e directory, no Playwright, no Cypress, no Jest for front-end code.

### What this means

The migration gate must be built before migration work starts. Without it, any change to the build pipeline has no automated way to confirm the output is functionally equivalent.

### Validation approach (three tiers)

**Pre-check before writing any baseline: confirm grunt is deterministic.**
Run grunt twice from clean, diff the two outputs. If they differ (timestamps in imagemin metadata, etc.) tier 1 is broken before it starts. This takes an hour and must happen first.

**Tier 1 — Module manifest diff (catches actual bundler migration failures)**

r.js emits a build manifest listing every module included per bundle. webpack produces the same via `stats.json`. Diff the *set of modules* before and after each change. This is the only reliable way to catch a silently dropped or duplicated module — which is the actual failure mode of a bundler migration. A ±5% size band would miss a dropped module in a 2 MB bundle; module manifest diff catches it deterministically.

**Tier 2 — Hash of stable static assets only**

SHA-hash only assets that are genuinely static: **images and fonts**. Not HTML, not JSON.
- HTML is not static — `grunt-inline` embeds minified JS/CSS into HTML after requirejs/terser run. Hashing HTML will false-fail on every minifier change.
- JSON goes through `grunt-json-minify` — key ordering/whitespace is tool-dependent.
- Images and fonts are byte-identical unless the image pipeline changes (4.3 is the only step that touches it — re-baseline after 4.3, see Step 4 ordering note).

**Tier 3 — File presence (hardest gate)**

Every expected output file must exist. A missing file is always wrong, regardless of content. This is a simple set diff of the file list.

**Layer 2 — Functional smoke test (manual, per editor, before merging)**

Document a minimal smoke test per editor:
- Editor loads without JS errors in the browser console
- File opens
- Basic editing works
- Save round-trip completes

This is manual for now. A minimal headless "editor opens, no console errors" check via Playwright is a low-cost addition once the first editor migrates — it catches the most common class of webpack migration failure (implicit global leakage from AMD code that relied on undeclared globals that webpack module-wrapping hides).

### What to do now

- [ ] **First**: run grunt twice, diff outputs — confirm determinism
- [ ] Write `build/scripts/baseline.js` (module manifest + image/font hashes + file list)
- [ ] Run a clean Grunt build to establish `baseline.json`
- [ ] Document the manual smoke test steps per editor
- [ ] Add baseline diff step to CI workflow

**Output:** `baseline.json` committed, smoke test steps documented. Nothing in Steps 2+ starts without this.

---

## Step 2 — Deep dive: unexpected side effects

Read everything we haven't read yet before writing any code:

- [ ] `web-apps/build/plugins/grunt-inline/` — document exactly what it inlines and where
- [ ] `web-apps/build/sprites/` (`iconsprite`) — document what it produces; confirm whether PNG sprites are still load-bearing or vestigial after the SVG migration
- [ ] `web-apps/build/theme.config.mjs` — read in full; confirm it is the shared config contract with the mobile webpack pipeline
- [ ] `web-apps/build/common.json` and all per-editor JSON configs — understand the full r.js config shape
- [ ] `vendor/framework7-react/build/webpack.config.js` — read the mobile webpack config in full; the desktop config must replicate its theme contract
- [ ] CI configuration — find it (GitHub Actions), understand build triggers and what currently runs in CI
- [ ] Check for any existing `grunt-contrib-watch` config — understand the current dev workflow

**Output:** a supplementary findings section appended to `build-system-report.md`.

---

## Step 3 — Branch

Create `build/webpack-migration` from current `main`.
All migration work lands on this branch. PRs are per-step, not per-phase.

---

## Step 4 — Low-hanging fruit (order matters)

Each item: change → build → diff → gate → commit.

**Ordering note:** 4.3 (imagemin replacement) changes image output bytes — do it *before* committing the baseline in Step 1, or explicitly re-baseline after. Similarly, 4.6 (version string injection) may affect hashed files — pin the version during baseline builds. The order below assumes 4.1 and 4.2 happen after baselining; 4.3 happens before or triggers a re-baseline.

**Step 4.0 is intentionally first** — it validates the baseline diff workflow with a known, predictable outcome (exactly 12 files removed) before any subsequent change adds uncertainty.

| # | Task | Risk | Notes |
|---|------|------|-------|
| 4.0 | **Remove IE 11 compat bundles** | Low | See detailed scope below |
| 4.1 | Delete Bower (`bower.json`, `.bowerrc`) | None | Confirmed dead; vendored packages in `web-apps/vendor/` |
| 4.2 | Remove `chcp` call (`Gruntfile.js:45`) | None | Windows codepage detection; dead on Linux |
| 4.3 | Replace `grunt-contrib-imagemin` with `sharp`/`svgo` CLI | Low | CVEs; do before baselining or re-baseline after — changes image bytes |
| 4.4 | Replace `grunt-text-replace` | Low | 5 lines of Node; token substitution is central to branding — test carefully |
| 4.5 | Fix Makefile branding defaults | None | `PUBLISHER_NAME`, `PUBLISHER_URL`, `DOCUMENT_ROOT` |
| 4.6 | ~~Move `sed` version patching into build-time injection~~ | — | **DROPPED** — phantom step. The only `sed` in the tree patches `server/Common/sources/commondefines.js` (DocService server), not web-apps. Web-apps already does version injection entirely in Grunt (`replace:writeVersion`). No web-apps sed exists to fold in. |
| 4.7 | Document multi-repo input dependencies + preflight assertion | None | Doc + executable `test -d` preflight; upgrade from prose — webpack won't warn on a missing path, it'll just emit a smaller bundle |

### Step 4.0 — Remove IE 11 compatibility bundles

**Rationale:** IE11 reached end-of-life June 2023. The current build produces `ie/app.js` + `ie/code.js` per editor (12 files total) via `terser:iecompat` and Babel transpilation. These are dead output that consumes one extra `terser` pass per editor and adds complexity with no benefit. The `--skip-babel` flag already exists in the codebase and suppresses this output when passed.

**Why now (before 4.1):** This step uses the baseline diff workflow for the first time, against a predictable outcome — exactly 12 files removed, module manifests for the main bundles unchanged. If the diff infrastructure has any bugs, they surface here on a safe change before they matter.

**Expected baseline diff result:**
- 12 `MISSING FILE` notices (ie/app.js + ie/code.js × 6 editors) — expected, not failures
- No `MODULES DROPPED` from any remaining bundle
- Re-baseline after this step so ie/ files are no longer in the reference

**Full scope of changes:**

*Build system (Gruntfile.js / common.json):*
- Remove `terser:iecompat` task config from Gruntfile.js (lines 684–694)
- Remove `babelTask` variable and its use in task registration (line 1003, and its spread into `deploy-app-main`)
- Remove `babel` task registration
- Remove `iecompat` copy task entry from `common.json` (line 97) — this copies `fix-ie-compat.js`
- Remove the `replace:writeVersion` ie/ conditional (lines 578–588, the `if !skip-babel` branch)
- Remove `--skip-babel` flag handling entirely (no longer needed; IE path is gone)

*Source HTML templates (*.html.deploy):*
All six editors have `isIEBrowser === true ? require(['ie/app']) : require(['app'])` or equivalent in their `index.html.deploy` and `index_loader.html.deploy` files. These need to collapse to the non-IE branch unconditionally:
- `apps/visioeditor/main/index.html.deploy` (line 509)
- `apps/visioeditor/main/index_loader.html.deploy` (line 339)
- `apps/pdfeditor/main/index.html.deploy` (line 408)
- `apps/pdfeditor/main/index_loader.html.deploy` (line 343)
- `apps/documenteditor/main/index.html.deploy` (line 439)
- `apps/documenteditor/main/index_loader.html.deploy` (line 344)
- `apps/documenteditor/forms/index.html.deploy` (line 276)
- `apps/presentationeditor/main/index.html.deploy` (line 459)
- `apps/presentationeditor/main/index_loader.html.deploy` (line 342)
- `apps/spreadsheeteditor/main/index.html.deploy` (line 520)
- `apps/spreadsheeteditor/main/index_loader.html.deploy` (line 343)
- `apps/spreadsheeteditor/main/index_internal.html.deploy` (line 289)

*Source app.js (per editor):*
Each editor's `app.js` has an `isIEBrowser` conditional for `code_path` that selects `ie/code` vs `code`. Remove the IE branch, unconditionally use `code`:
- `apps/visioeditor/main/app.js` (line 167)
- `apps/pdfeditor/main/app.js` (line 199)
- `apps/documenteditor/forms/app.js` (line 157)
- `apps/documenteditor/main/app.js` (line 206)
- `apps/spreadsheeteditor/main/app.js` (line 203)
- `apps/presentationeditor/main/app.js` (line 203)

*Source file (optional, confirm first):*
- `apps/common/main/lib/util/fix-ie-compat.js` — confirm it is only referenced by the `iecompat` copy task and not imported by any AMD module before deleting.

**Advisor-reviewed (claude-opus-4-8, 2026-06-13):**

**Makefile correction:** `develop/setup/Makefile:63` is the `web-apps-dev` target — it passes `--skip-babel`. The `web-apps` (production) target at line 55 does NOT pass `--skip-babel`. **IE bundles ARE in production today.** The `baseline.json` captured during Step 1 is correct and includes the ie/ bundles.

After the code change: re-baseline against a fresh grunt build (production-equivalent, no `--skip-babel` since the flag will be gone). Expected diff: 12 `ie/` bundle files missing + `fix-ie-compat.js` missing + HTML content changes on the built index.html files (not failures — all expected).

**Corrected full scope of changes (post-advisor review):**

*Build system:*
- `build/Gruntfile.js`: Remove `terser:iecompat` (lines 684–693), `babel:` task config (lines 697–705), `babelTask` var + spread (lines 1003, 1023), `replace:writeVersion` ie/ conditional (lines 579–584)
- `build/Gruntfile.js:286`: Remove `grunt.loadNpmTasks('grunt-babel')` — **missed in original scope**
- `build/appforms.js:167`: Remove `babelTask` / `iecompat` declaration
- `build/common.json:97`: Remove `iecompat` copy task entry
- `develop/setup/Makefile:63`: Remove `--skip-babel` from `web-apps-dev` target (keep in lockstep with Gruntfile)
- `web-apps/Readme.md`: Remove `--skip-babel` documentation (lines 101, 109)

*Source HTML templates (collapse isIEBrowser conditionals — require(['app']) unconditionally):*
Same 12 files listed above. Additionally, each `index.html.deploy` block injects **two** IE scripts, not one:
```js
isIEBrowser === true &&
    (document.write('<script src=".../fix-ie-compat.js"><\/script>'),
     document.write('<script src=".../sdkjs/vendor/string.js"><\/script>'));
```
Remove the entire `isIEBrowser && (...)` block. **`string.js` injection was missed in original scope.**

Also: `apps/presentationeditor/main/index.reporter.html.deploy` has an IE detection block — remove it.

*Source app.js (6 editors):* Same as before — unconditionally use non-IE `code_path`.

*Runtime `isIEBrowser` guards in utility files — **do NOT touch in this PR**:*
`apps/common/main/lib/util/htmlutils.js:57` and `apps/common/main/lib/util/themeinit.js:42-43` check `window.isIEBrowser` for RTL layout and theme defaults. These are runtime feature guards on a global, not dead code. They safely evaluate the non-IE branch on all modern browsers. Simplifying these ternaries is a semantically-loaded cleanup that touches RTL and theming — keep it in a separate, bisectable follow-up.

*`embed/` editors:* `apps/{documenteditor,spreadsheeteditor,presentationeditor,visioeditor}/embed/index.html.deploy` have IE detection blocks rendering a "browser not supported" mask. They do not produce `ie/` bundles, so this PR won't break them. Leave them; note they exist as inconsistent leftovers for a follow-up.

*Committed built `.html` files:* `git ls-files` shows ~15 built `apps/**/index.html` alongside the `.deploy` sources. They're artifacts — the build regenerates them via `copy:indexhtml` + `replace:indexhtml`. Run `grunt clean` before rebuilding to avoid serving stale files from the `$BUILD_ROOT`.

*Delete:* `apps/common/main/lib/util/fix-ie-compat.js` — standalone polyfill, not an AMD module. Delete LAST, after removing both the copy task in common.json and the `document.write` references in HTML deploy files.

**Execution order (IMPORTANT — wrong order produces 404s):**
1. HTML deploy files + app.js source (remove ie/ require branches and script injections)
2. Gruntfile.js + appforms.js + common.json (remove tasks and copy config)
3. Delete `fix-ie-compat.js`
4. Makefile + Readme cleanup

Rationale: if the Grunt tasks are removed first but HTML still has `require(['ie/app'])`, any build produces HTML that references bundles that no longer exist. Source-before-Gruntfile ensures the outputs are always consistent.

---

## Step 5 — webpack 5 consolidation (Phase 1)

Replace the Grunt task zoo with webpack loaders/plugins. Replace r.js with webpack's native AMD handling. One editor at a time, using e2e as the gate per editor.

### Baseline gate strategy change (Opus review, v6)

**The module-manifest diff gate is wrong for a bundler swap.** r.js and webpack will never produce identical module manifests or byte-identical bundles — the gate will be red by construction after migrating any editor. For migrated editors, replace module-manifest diff with:
1. **Output-filename presence gate** — every expected `app.js`, `code.js`, locale JSON, CSS, and image file must be present at the exact path the packaging contract expects.
2. **Behavioural smoke test** — editor boots, controllers register by name, `code.js` loads on demand (see smoke-test.md).

The module-manifest/hash gate remains valid for Grunt-only editors (not yet migrated) — it catches regressions in the unchanged editors during the dual-build window.

### Pre-work (do once)

- [ ] **Audit named `define()`**: grep for `define('name',` (named modules) across all editors. These are the AMD ids that webpack ignores (webpack resolves by file path, not by the explicit id string). Every named module must become a `resolve.alias` entry mapping `id → file path`. Failure to do this produces *silent* broken links — `require('name')` just resolves nothing. Expected count: ~19 based on advisor scan.

- [ ] **`text!` plugin via asset/source**: Do NOT write a custom loader. Add a `NormalModuleReplacementPlugin` that strips the `text!` prefix and rewrites the request, then add a rule `type: 'asset/source'` matching `.html` / `.template` extensions. The replacer must anchor on the AMD dependency string — a raw text scan will false-fire on `text!==false` (that is `!==` operator, not a plugin prefix). `.template` has no webpack-recognized extension by default — add it explicitly to the rule.

- [ ] **`code.js` as second webpack entry with `dependOn`**: `app_pack.js` is a bare `require([47 modules])` side-effect preload — no `define()`, no return value. DocumentServer's outer page loads `code.js` *after* `app.js`. Replicate as a second webpack entry: `entry: { app: './app.js', code: { import: './app_pack.js', dependOn: 'app' } }`. The `dependOn: 'app'` flag shares the single runtime chunk, so the `define()`-registered classes in `app.js` are visible to `code.js` without double-bundling. This is the most likely thing to silently break — validate it first in the spike.

- [ ] **Translate r.js `paths` to `resolve.alias`** — including the 19 named-define ids found in the audit above. `empty:` paths → webpack `externals`.

- [ ] **Translate r.js `shim` to `ProvidePlugin`** — Backbone/Underscore/perfectscrollbar global side effects; use `ProvidePlugin` for globals (`_`, `Backbone`); use `imports-loader` for dependency-ordering shimmed non-AMD libs.

- [ ] **Strip `require.config()` from HTML templates** — do NOT leave them. They will throw `require is not defined` at execution time when require.js is absent. The template must be cleaned during the migration of each editor. Move all alias truth into the webpack config.

- [ ] **AGPL license header preservation in TerserPlugin**: webpack's default Terser pass strips all comments, including Ascensio AGPL headers. This is a compliance regression the gate will not catch. Add `extractComments: false` and preserve license comments:
  ```js
  new TerserPlugin({
      extractComments: false,
      terserOptions: { format: { comments: /AGPL|Copyright|License/i } },
  })
  ```
  Verify the header survives in the emitted bundle before shipping any editor.

- [ ] **`_themVal()` → DefinePlugin: extend `theme.config.mjs`, do NOT reinvent.** `theme.config.mjs:themeDefines()` is the mobile mirror — it is already half-built. The desktop migration must extend it. Key delta: ~10 more replacement keys than mobile (`HELP_CENTER_WEB_*`, `SUGGEST_URL`, `API_URL_EDITING_CALLBACK`, `COEDITING_DESKTOP`, `PLUGIN_LINK*`, `DEFAULT_LANG`, `APP_COPYRIGHT` banner). List the full delta before writing any DefinePlugin config.

- [ ] **`{{PRODUCT_VERSION}}` injection** — `replace:writeVersion` in Grunt injects `version.build` into compiled JS. webpack must replicate via `DefinePlugin` or `string-replace-loader`. Add alongside `themeDefines()`.

- [ ] **Replicate the addon merge system** as a webpack config merge utility.

- [ ] **grunt-inline answer — keep as post-webpack pass (option a)**. Do NOT use HtmlWebpackPlugin `templateParameters` + `fs.readFileSync` for the SVG sprite. That approach inlines the SVG at config-eval time (not watched), loses grunt-inline's relative-path resolution, and adds migration risk. Keep grunt-inline as a thin post-webpack HTML pass for SVG sprite injection only — it is a 1:1 behavior match against the baseline gate.

- [ ] **Three HTML entry points per editor must stay in sync** — `index.html` (dev), `index_loader.html`, `index.html.deploy` (prod, processed by grunt-inline for `<inline>` SVG sprites). SVG sprite includes break if any of the three is missed.

- [ ] **Output path contract** — webpack output filenames must match what DocumentServer packaging expects (`code.js`, `app.js` per editor). No content-hashed filenames.

- [ ] **Baseline vs prod-bake gap** — baseline captured with `--skip-imagemin`. x86 prod bake Dockerfile runs imagemin. The gate must not run against prod bake output. Documented in `docs/multi-repo-deps.md`.

- [ ] **Spike on visioeditor** — validate: (1) two-entry `app.js`/`code.js` split with `dependOn` produces a working editor, (2) named-define resolution for found named modules, (3) `text!`→asset/source rewrite works across real templates, (4) output filenames match packaging contract, (5) green baseline file-presence gate. Note explicitly: a clean visioeditor spike does NOT de-risk documenteditor's `COEDITING_DESKTOP`/plugin breadth — those paths need separate validation.

**Per-editor migration order** (safest first):
1. `visioeditor` — newest, smallest, fewest legacy patterns, genuinely standalone
2. `pdfeditor`
3. `appforms` — **not** first despite being small; it builds from `apps/documenteditor/forms` (inside the documenteditor family), so it's a documenteditor rehearsal, not a simple standalone
4. `presentationeditor`
5. `spreadsheeteditor`
6. `documenteditor` (most complex — last)

Each editor: add webpack config → build → structural diff → e2e green → commit → move to next.

**Desktop build mode:** `--desktop=true` becomes a webpack `--env desktop` flag.

---

## Step 5.x — Remove grunt (endgame)

Gate: all five non-desktop grunt responsibilities must be replaced before this step.

**Precondition checklist:**
- [ ] SVG sprite *injection* replaced (inline SVG content into HTML, all editors × both HTML entry points — sprite *generation* via `sprites.sh` already standalone)
- [ ] `deploy-theme-images` replaced (themed images into mobile dirs)
- [ ] embed builds replaced (per editor: terser + less + copy + inline)
- [ ] forms build replaced (`appforms.js` / documenteditor/forms)
- [ ] mobile build direct-call wired (CI + Makefile: `npm i` in vendor/framework7-react, then `node build/build.js` with `TARGET_EDITOR` mapping)

**When preconditions are met:**
1. Delete `build/Gruntfile.js` and `build/appforms.js`
2. Remove grunt devDeps from `build/package.json`
3. Remove grunt step from `.github/workflows/build.yml`
4. Remove grunt call from `develop/setup/Makefile` (DocumentServer repo)
5. Run CI — confirm all editors build, all outputs present
6. Run manual smoke test per editor

---

## Step 6 — webpack 5 consolidation (Phase 2)

Opportunistic AMD → ESM conversion. No freeze, no big-bang.

- New code is written as ESM from this point forward
- Existing modules convert to ESM when touched for other work
- Track conversion progress; aim for majority-ESM before considering Phase 3

---

## Step 7 — Phase 3 (future, optional)

Once majority-ESM: evaluate Rspack (drop-in webpack-API Rust rewrite with AMD support) for build speed. Or migrate to Vite — at majority-ESM, that becomes a config change.

**sdkjs/Closure Compiler is out of scope for all phases.**

---

## Risks

| Risk | Mitigation |
|------|-----------|
| Branch lifetime — `build/webpack-migration` rotting against main across six editor migrations | Merge per-editor behind a dual-build flag; don't accumulate six editors on one long-running branch |
| Dual-build window — while some editors are grunt and some are webpack, CI must run both pipelines | Budget for this; document it explicitly in CI config |
| Vendor packages in `web-apps/vendor/` with non-standard module formats | Each may need bespoke shims; 557 AMD files is the known surface, vendor is the unknown one — audit during Step 2 |
| Implicit global leakage — AMD code relying on undeclared globals breaks when webpack wraps modules | Only the smoke test catches this; the minimal headless Playwright check is the safety net |
| Named `define('id', …)` modules — webpack ignores explicit AMD module ids, resolves by file path | Audit all named defines; add each as a `resolve.alias` entry before migrating any editor |
| AGPL license headers stripped by Terser | Configure `TerserPlugin` with `extractComments:false` and preserve AGPL/Copyright comments; verify in emitted bundle |
| Baseline gate wrong for bundler swap — module manifests will never match | Switch migrated-editor gate to file-presence + smoke test; keep module-manifest gate only for unmigrated editors |
| `require([...], cb)` becomes async chunk boundary in webpack by default | This changes runtime load order; test that dialog modules still register correctly after migration |

---

## Step 5 — findings so far

### What was built

Four webpack configs created and tested:

| Editor | Config | Build validates | Runtime validates |
|--------|--------|----------------|-------------------|
| visioeditor | `build/webpack.visioeditor.js` | ✓ | — (not runtime-testable locally) |
| documenteditor | `build/webpack.documenteditor.js` | ✓ | ✓ (editor loads, SDK resolves, Backbone boots) |
| spreadsheeteditor | `build/webpack.spreadsheeteditor.js` | ✓ | pending user test |
| presentationeditor | `build/webpack.presentationeditor.js` | ✓ | pending user test |
| pdfeditor | `build/webpack.pdfeditor.mjs` | ✓ | pending user test |

### Key technical decisions

**AMD externals approach:** `externalsType: 'amd'` (global) + per-entry `library: { type: 'amd', name: 'app' }`.
- Without both, webpack compiles externals to `module.exports = void 0`
- `externalsType: 'amd'` generates `require('sdk')` AMD stubs
- Per-entry library wraps output in `define("app", ["api","sdk"], factory)` so require.js loads externals before the factory runs
- Result: `define("app", ["api","sdk"], factory)` with externals properly wired

**HTML pre-config (discovered requirement):** The `require.config({paths, shim})` call inside app.js source is a no-op in webpack (it calls `__webpack_require__.config()`, not require.js). But require.js needs those paths to resolve externals declared in the `define()` deps array — specifically the sdk path and the shim for sdk's own dependencies. Fix: move `baseUrl`, `paths`, and `shim.sdk` into the `var require = {...}` pre-config block in `index.html.deploy`. This block IS processed by require.js before module loading begins.

**`asyncChunks: false`:** AMD `require([...], callback)` in source becomes async chunk boundaries by default. DocumentServer packages by exact filename — stray chunk files would be silently missing. `asyncChunks: false` merges all lazy AMD requires into the parent entry.

**noParse for locale.js:** `require([polyfill], cb)` inside a plain IIFE body (not inside `define()`) crashes the AMD parser. The polyfill path is dead in modern browsers. `noParse` bypasses AMD parsing for this file.

**`parser: { amd: false }` for LaunchController.js:** The 3-arg require.js form `require({waitSeconds:0}, dynamicArray, cb)` is unrecognisable to webpack's AMD parser. In webpack world `code.js` is loaded separately; the dynamic dispatch is superseded.

**Named define audit:** Zero named `define('id', ...)` calls found in the bundled app source for all four editors. (Named defines exist only in `vendor/` test fixtures and `help/` pages — out of scope.) The plan's concern about ~19 named defines does not apply.

**Build timing (documenteditor, Mac, warm):**
- webpack: 3.3 s wall clock (LESS + bundle + minification + locale copy)
- grunt r.js pipeline (--skip-imagemin --skip-sprites): 4.8 s wall clock (LESS + r.js x2 + terser x2 + concat + copy + svgmin + replace + inline + json-minify)
- webpack ~1.5× faster. Gain will compound when editors build in parallel (grunt is serial).

### Deviations from the plan

1. **Editor order** — plan said: visioeditor → pdfeditor → appforms → presentationeditor → spreadsheeteditor → documenteditor. Actual: visioeditor (spike, not runtime-testable) → documenteditor (runtime-validated) → spreadsheeteditor + presentationeditor. documenteditor was tackled second because it is in Euro Office and gives a real runtime gate. pdfeditor and appforms/forms not yet done.

2. **require.config() handling** — plan said "strip require.config() from HTML templates". Actual: kept require.js as the AMD loader; moved paths/shim from the source `require.config()` call into the HTML `var require = {...}` pre-config. require.js remains the loader for the SDK and other AMD externals. The source `require.config()` call in app.js remains in the bundle as a no-op — it is harmless, not a throw.

3. **shim → ProvidePlugin** — plan pre-work said to use ProvidePlugin for Backbone/Underscore globals. Actual: resolve.alias to vendored files works without ProvidePlugin; Backbone and Underscore are bundled directly. No globals leak. ProvidePlugin not needed for Phase 1.

### Advisor findings (v7, 2026-06-14)

**CRITICAL — `{{TOKEN}}` string literals cannot be replaced by DefinePlugin.**

Desktop source uses mustache-style `{{TOKEN}}` delimiters embedded _inside string literals_:
```js
new URL('{{HELP_CENTER_WEB_DE}}')
publishername: '{{PUBLISHER_NAME}}'
window.open('{{SUGGEST_URL}}', ...)
```
`DefinePlugin` replaces free **identifier nodes** in the AST; it cannot rewrite text inside a string literal. All three bundles currently ship with unreplaced `{{HELP_CENTER_WEB_*}}`, `{{PUBLISHER_NAME}}`, `{{SUGGEST_URL}}` etc. baked in verbatim. documenteditor "runtime validated" because boot does not exercise these paths — they surface in the About box, help-center links, and feedback URL.

`{{PRODUCT_VERSION}}` has the same bug: 3 mustache-form occurrences exist alongside any `__PRODUCT_VERSION__` bare identifier. Only the bare identifier form is caught by DefinePlugin.

**Correct fix:** string-replace loader pass over emitted source keyed on the same `{{TOKEN}} → value` table that Grunt already uses (`global.jsreplacements` in Gruntfile). Factor that table into `theme.config.mjs` as a single exported list, consumed by **both** Gruntfile (no change to existing key) and webpack (via `string-replace-loader`). Do NOT extend `themeDefines()` for these — DefinePlugin is the wrong tool for this job.

**mangle:false size impact:** Disabling Terser mangling inflates bundle size relative to the grunt baseline. Intentional (safe choice against AMD/Backbone string-name coupling), but the CI baseline-diff size gate must account for this before enabling mangle:true in a follow-up.

**Config triplication:** The three editor configs (doc/spreadsheet/presentation) are byte-identical after name substitution. Consider a `webpack.editor.factory.mjs(editorName, sdkPath)` that all four thin configs call — cheap now, saves painful divergence later.

**HTML pre-config sync risk:** The `var require = {paths, shim}` block in `index.html.deploy` and the `externals` object in the webpack config must stay in sync. Long-term end state: drop require.js, load externals as plain `<script>` globals and switch `externalsType` to `'window'`. Not worth doing mid-migration — but mitigate the sync risk by generating the HTML `paths` block from the webpack externals object (HtmlWebpackPlugin template var, or a shared constant in `theme.config.mjs`). 

**Editor order / pdfeditor:** pdfeditor IS in Euro Office's active editor set — webpack config and CI gating needed. The three validated editors (doc/spreadsheet/presentation) went into CI first; pdfeditor is next. appforms (lives in `apps/documenteditor/forms`) is not confirmed in Euro Office scope — keep deferred until explicitly confirmed.

### Corrected priorities (post-v7 advisor review)

1. ~~**Token substitution fix**~~ ✓ **DONE** — `string-replace-loader` + `themeReplacements()` in `theme.config.mjs`. 19 tokens (including `{{HELP_CENTER_WEB_VE}}` discovered during scan). Gated by `--scan-tokens` in `baseline.js`. Commit `36d3060d93`.
2. ~~**`index_loader.html.deploy` pre-config**~~ ✓ **DONE** — applied to doc/spreadsheet/presentation. Commit `b8ee169103`.
3. ~~**CI integration** (doc/spreadsheet/presentation)~~ ✓ **DONE** — webpack runs for all four configs in CI. Commits `de927813df`, `f752f091f6`.
4. **grunt-inline SVG sprite replacement** — last grunt dependency in the runtime path; needed before grunt can be retired for the three editors. **← NEXT**
5. **pdfeditor** — IN SCOPE. webpack config + `index_loader.html.deploy` pre-config + CI integration pending. (appforms deferred — not confirmed in Euro Office scope.)

### Step 5 completed work (2026-06-14)

| Item | Commit | Notes |
|------|--------|-------|
| `--scan-tokens` mode added to `baseline.js` | `d22a89a3` | Scans .js/.css output for unreplaced `{{TOKEN}}`; exits 1 on any match. `.json` excluded (locale runtime interpolation). |
| `themeReplacements(productVersion)` added to `theme.config.mjs` | `36d3060d` | 19 entries; string-replace-loader 'multiple' format; covers all tokens found in scan incl. `{{HELP_CENTER_WEB_VE}}` (visioeditor-specific, absent from Gruntfile jsreplacements). |
| `string-replace-loader` rule in all 4 webpack configs | `36d3060d` | Applies to `.js` files under `APPS_ROOT`. Import chain: webpack config → `themeReplacements(productVersion)` → `theme.config.mjs`. |
| `index_loader.html.deploy` pre-config | `b8ee169103` | `var require = {baseUrl, paths, shim}` added to doc/spreadsheet/presentation. Visioeditor not in Euro Office; its index_loader left unchanged. |

### Still pending from Step 5

- [x] **CI integration** — ✓ DONE. webpack runs for all four editors (doc/spreadsheet/presentation/visioeditor) in parallel in CI. Commits `de927813df`, `f752f091f6`.
- [x] **grunt-inline SVG sprite *injection* replacement** — `build/scripts/inline-svgs.js` written. Mirrors grunt-inline's exact `<inline>` regex; globs `*.html` in each editor's main output dir (covers `index.html` + `index_loader.html` + `index_internal.html`). Removed `'inline'` from `deploy-app-main` in Gruntfile. CI step added after grunt, before webpack. Key learning: `index_loader.html.deploy` has `<inline>` tags in all 5 editors; `spreadsheeteditor/main/index_internal.html.deploy` has 2; `uglify: true` was a no-op for SVG (grunt-inline never minified the `<inline>` branch). Makefile (DocumentServer repo) still needs `node build/scripts/inline-svgs.js` added after the grunt step — separate PR.
- [ ] **`deploy-theme-images` replacement** — webpack configs only copy `locale`. Grunt's `deploy-theme-images` copies themed images into `apps/<editor>/mobile/resources/img`. Retiring grunt without replacing this drops themed images for both desktop and mobile.
- [ ] **embed builds** — `deploy-app-embed` (terser + less + copy + inline) for doc/spreadsheet/presentation/visio. pdfeditor has no embed. Not currently webpack-owned.
- [x] **pdfeditor webpack config** — `build/webpack.pdfeditor.mjs` created, builds clean (2488 ms). `index.html.deploy` + `index_loader.html.deploy` pre-config blocks added. CI integration added to `build.yml`. Runtime validation pending (open in Euro Office). Note: pdfeditor has no embed build.
- [ ] **appforms / forms webpack config** — deferred; not confirmed in Euro Office scope. Lives in `appforms.js`, pulled in via documenteditor's deploy list.
- [ ] **Config factory** — `webpack.editor.factory.mjs(editorName, sdkPath)` to eliminate config triplication. Low urgency while configs are stable; high urgency if any config diverges.
- [ ] **`mangle: false` revisit** — intentional safety choice; bundle sizes inflate vs grunt baseline. Safe re-enable path: `mangle: { keep_classnames: true, keep_fnames: true }` — bare `mangle:true` breaks Backbone controllers registered by constructor name. Revisit after CI baseline is established.

---

## Gotchas (accumulated during Step 5)

**requirejs `baseUrl` trap** — `require(['app'])` with `baseUrl: "../../"` resolves to `/web-apps/apps/app.js` (404). The webpack bundle registers as `define("app", ...)` but require.js looks for it as a file path, not by AMD name. Fix: add `app: "editorName/main/app"` to the `var require = {paths}` pre-config in each editor's `index.html.deploy` and `index_loader.html.deploy`. Fixed: commit `6f52b4460c`.

**Service worker caches `app.js`** — `document_editor_service_worker.js` (lives in `sdkjs/common/serviceworker/`, deployed to `sdkjs/deploy/sdkjs/common/serviceworker/`) aggressively caches editor assets. Testing a new webpack build without clearing the SW cache silently serves an older build. Always test in a private/incognito window on first load, or hard-reset the SW cache in devtools. Confirmed real — appears verbatim in browser console stack traces.

**`grunt-inline` processes inside HTML comments** — `<!--<inline src="..." />-->` still triggers grunt-inline's file lookup even inside a comment. Dead inline references must be deleted entirely, not commented out. Fixed: removed dead `buttons.svg` reference from `spreadsheeteditor/main/index_internal.html.deploy` (commit `892d946c21`).

**webpack `BUILD_ROOT` env var** — webpack configs read `process.env.BUILD_ROOT` and default to `../deploy` (relative to `build/`). Inside the `eo` container `BUILD_ROOT=/var/www/onlyoffice/documentserver`. All configs already implement this. Without `BUILD_ROOT` set, webpack outputs to the local `deploy/` dir which is not where the container's nginx serves from.

**Makefile lives in a different repo** — `DocumentServer/develop/setup/Makefile` is in the `DocumentServer` repo, not `web-apps`. Any change to the local dev build sequence requires a PR in that repo. CI and the Makefile are separate and must be kept in sync manually.

**grunt must run before webpack — ordering is load-bearing** — both grunt and webpack write to the same `BUILD_ROOT/web-apps/apps/<editor>/main/`. Webpack configs all have `output.clean: false`. CI and Makefile both order grunt→webpack. Reversing the order, or dropping grunt for an editor without a clean step, leaves stale r.js-built `app.js`/`code.js` on disk.

**Mobile invocation is `node build/build.js`, not a raw webpack call** — `vendor/framework7-react/build/build.js` does `rimraf ./www/` before building and maps `TARGET_EDITOR` (word/cell/slide/visio). Calling the webpack config directly skips the clean and the env mapping. The `exec:webpack_install` step (`npm i` in `vendor/framework7-react`) must also be preserved.

**`deploy-theme-images` is the only thing copying themed images into mobile dirs** — webpack configs copy `locale` only. Retiring grunt without replacing `deploy-theme-images` silently drops themed images from both desktop and mobile.

**HTML pre-config blocks must track `externals` by hand** — four `index.html.deploy` + four `index_loader.html.deploy` carry `var require = {paths, shim}` that must match each config's `externals` object. Drift between them produces a silent runtime 404 that the file-presence gate cannot catch. This is already the root cause of commit `6f52b4460c`. CI and the Makefile both independently list all four editor configs — a further three-way sync surface once mobile/embed get direct calls.

**`drop_console` — webpack drops console logs, grunt does not** — webpack sets `drop_console: env === 'production'`; since `NODE_ENV` is not set in CI, `env` defaults to `'production'`, so webpack strips all `console.*` calls in every CI build. Grunt uses `compress: true` with no `drop_console` — terser default is `false`, so grunt preserves console calls. This is a silent production behaviour change introduced by the webpack migration. Decide whether to align (add `drop_console: true` to grunt, or remove it from webpack) before declaring output equivalent.

**`asyncChunks: false` is required** — AMD `require([...], callback)` in source becomes an async chunk boundary by default in webpack. DocumentServer packages by exact filename; stray chunk files are silently missing at runtime. `asyncChunks: false` merges all lazy AMD requires into the parent entry.

**`noParse` for `locale.js`** — `require([polyfill], cb)` inside a plain IIFE (not inside `define()`) crashes webpack's AMD parser. The polyfill path is dead on modern browsers. `noParse` bypasses AMD parsing for this file.

**`parser: { amd: false }` for `LaunchController.js`** — the 3-arg require.js form `require({waitSeconds:0}, dynamicArray, cb)` is unrecognisable to webpack's AMD parser. In the webpack build, `code.js` is loaded separately; the dynamic dispatch is superseded. Disable AMD parsing for this file only.

---

## Step 6 — CI integration scope

### Current CI workflow (`.github/workflows/build.yml`)

```
checkout web-apps + sdkjs → Node 20 → sprites.sh → npm install →
translate merge/check → grunt --skip-imagemin → baseline diff
```

One job, self-hosted runner, no parallelism. Grunt builds all six editors serially.

### Design principle: CI is the project's build pipeline, not migration scaffolding

Migration validation tools (`baseline.js`, `--scan-tokens`, baseline diffs) are local migration aids — they run during development work, not in CI. They will be removed from the repo when migration is complete. Adding them to CI creates debt that needs cleaning up later.

**CI contains only what the project needs permanently:**

```
... (existing steps unchanged) ...

- Run grunt (all editors — unchanged)

- Run webpack (doc + spreadsheet + presentation + pdfeditor, in parallel)
    cd web-apps/build
    THEME=euro-office npx webpack --config webpack.documenteditor.mjs     &
    THEME=euro-office npx webpack --config webpack.spreadsheeteditor.mjs  &
    THEME=euro-office npx webpack --config webpack.presentationeditor.mjs &
    THEME=euro-office npx webpack --config webpack.pdfeditor.mjs          &
    wait
```

The webpack step fails CI if any editor build errors. That is the gate. No baseline diffs, no token scans in CI.

### Open questions before implementation

1. **Output path mismatch — confirm before writing any CI step.** Both grunt and webpack resolve `BUILD_ROOT` to `path.resolve('../deploy')` from `web-apps/build/` → `{workspace}/web-apps/deploy/`. But the baseline diff step runs from the workspace root and reads `deploy/web-apps/apps` → `{workspace}/deploy/web-apps/apps/`. These are different paths. The mismatch may mean the existing baseline diff step has never actually compared against a real grunt-built tree (baseline.json was captured inside the `eo` container, not from a CI run). **Pre-flight: trigger CI on this branch and confirm the baseline diff step passes. If it fails on path, fix the baseline diff path first (`web-apps/deploy/web-apps/apps`), then re-capture baseline.json from a clean CI grunt run before adding webpack steps.**

2. **Baseline re-capture** — `baseline.json` was captured from grunt output. After the first successful dual-build CI run, baseline.json must be re-captured from the mixed output (grunt for visioeditor/appforms, webpack for doc/spreadsheet/presentation/pdfeditor) and committed. Module manifests for the webpack editors will shrink to just `["app"]` — that is expected and not a failure.

3. **Build timing capture** — wrap both the grunt step and webpack step with `time` (or capture `date +%s` before/after) to get the CI performance headline for the grunt-vs-webpack comparison.

### Lessons from Step 5 that apply directly

- `THEME=euro-office` is required; webpack reads `theme/euro-office/meta/config.json` at config-eval time.
- All four webpack builds are independent — they can run in parallel in CI (`& wait`) unlike grunt which is serial. This is the headline performance improvement.
- `--scan-tokens` is already written and gated to exit 1 on failure — just add it as a CI step.
- No `sdkjs` checkout needed for webpack builds specifically — webpack treats sdkjs as an external. But sdkjs is still needed while grunt builds any remaining editors (visioeditor, appforms).

---

## Review log

| Version | Date | Author | Changes |
|---------|------|--------|---------|
| v1 | 2026-06-12 | Claude Sonnet | Initial plan |
| v2 | 2026-06-12 | Fable (claude-fable-5) | HTML/JSON removed from hash tier (grunt-inline embeds minified JS); size-bands replaced with module-manifest diff; grunt determinism pre-check added; imagemin ordering note; three Step 5 pre-work gaps added (entry-point HTML, output path contract, grunt-inline answer); appforms moved out of first position (builds from documenteditor/forms); visioeditor confirmed first; risks table added |
| v3 | 2026-06-13 | Claude Sonnet | Added Step 4.0 IE removal with full scope (12 html.deploy files, 6 app.js files, build/appforms.js, Makefile ref); documented key finding that develop/setup/Makefile already passes --skip-babel (IE bundles absent from production today) |
| v4 | 2026-06-13 | Opus (claude-opus-4-8) | Corrected: --skip-babel is web-apps-dev only; production build DOES produce IE bundles; IE bundles ARE in production. Added: missing grunt.loadNpmTasks('grunt-babel') removal, string.js injection in HTML blocks (not just fix-ie-compat.js), presentationeditor/reporter.html.deploy, embed editor awareness. Clarified: leave htmlutils/themeinit isIEBrowser guards (not dead code, runtime guards). Added explicit execution order (HTML source first, then Gruntfile, then delete fix-ie-compat.js). |
| v5 | 2026-06-13 | Opus (claude-opus-4-8) | Dropped 4.6 (phantom — no web-apps sed; server sed is unrelated). Deferred 4.3 (imagemin never runs; document prod-bake baseline gap). Upgraded 4.7 to doc+preflight. Step 5 pre-work corrections: extend theme.config.mjs/themeDefines() not reinvent; add {{PRODUCT_VERSION}} DefinePlugin item; three HTML entry points per editor (SVG sprite trap); content-hash filename trap from mobile pipeline; prod-bake vs baseline gap. |
| v6 | 2026-06-13 | Opus (claude-opus-4-8) | Critical Step 5 corrections: (1) Baseline gate is wrong for a bundler swap — module manifests will never match r.js output; switch to file-presence + smoke test for migrated editors. (2) ~19 named `define('id',…)` modules — webpack ignores AMD ids, must audit and add each as resolve.alias. (3) code.js must be second webpack entry with `dependOn: 'app'` not a dynamic import. (4) require.config() in HTML must be stripped (throws if require.js absent). (5) grunt-inline answer reversed: keep as post-webpack pass (option a), NOT HtmlWebpackPlugin templateParameters. (6) AGPL headers stripped by Terser — must configure TerserPlugin to preserve. Added four new risk rows. |
| v7 | 2026-06-14 | Opus (claude-opus-4-8) | Critical: DefinePlugin cannot fix desktop brand tokens — source uses `{{TOKEN}}` inside string literals, not bare `__TOKEN__` identifiers; all three webpack bundles currently ship unreplaced brand strings. Fix: `string-replace-loader` with shared map in `theme.config.mjs` (same source as `global.jsreplacements`). `{{PRODUCT_VERSION}}` has the same bug (3 mustache occurrences). Corrected priority order: (1) token fix first — blocks trustworthy CI baseline, (2) `index_loader.html.deploy` pre-config NOW not deferred, (3) CI for 3 editors, (4) grunt-inline SVG, (5) pdfeditor/appforms deferred (not in Euro Office). Named define audit result validated correct (zero named defines in editor source). Additional flags: mangle:false inflates bundles vs grunt baseline; config triplication across 3 editor configs; HTML pre-config sync risk — mitigate by generating paths from shared constant. |
| v8 | 2026-06-14 | Claude Sonnet | Completed v7 priorities 1 and 2. Token fix: `themeReplacements(productVersion)` exported from `theme.config.mjs`; 19-entry string-replace-loader map; scan discovered `{{HELP_CENTER_WEB_VE}}` (visioeditor-specific, absent from Gruntfile). `--scan-tokens` mode added to `baseline.js` as ongoing gate. `index_loader.html.deploy` pre-config applied to doc/spreadsheet/presentation. Added topology reference at `.claude/migration-topology.md`. Next: CI integration scoping. |
| v9 | 2026-06-14 | Opus (claude-opus-4-8) | Plan had drifted from repo state. Key corrections: (1) CI already runs webpack for all four editors including visioeditor — CI integration is done, not pending. (2) visioeditor is fully in scope (in CI + Makefile), not deferred. (3) grunt retirement table was missing three rows: `deploy-theme-images` (only thing copying theme images into mobile dirs), `deploy-app-embed` (embed builds, per editor, in grunt default), `deploy-app-forms` (in appforms.js). (4) Mobile invocation is `node build/build.js` + `npm install` in vendor dir, not a direct webpack config call — `rimraf` clean and TARGET_EDITOR mapping must be preserved. (5) No "delete grunt" endgame step existed — added Step 5.x with precondition checklist. (6) `mangle:true` revisit: safe path is `keep_classnames:true, keep_fnames:true`, not bare enable. (7) Added gotchas: grunt→webpack ordering is load-bearing (`clean:false`), theme-images gap, mobile invocation detail, HTML pre-config three-way sync risk, `drop_console` parity check. |

---

## Current status

- [x] Build system analysis complete (`build-system-report.md`)
- [x] Fable review of analysis incorporated (v2 of report)
- [x] Migration plan written
- [x] Fable review of plan incorporated (v2 of plan)
- [x] **Step 1.1 — build environment confirmed** (Node 18 + grunt in `eo` container; Node 20 in bake Dockerfile; local Node 25 — do not switch)
- [x] **Step 1.2 — determinism check** — skipped for now; deployed files in `eo` used as baseline instead of re-running grunt (rebuilding the image is the only way to run grunt twice cleanly; deferred)
- [x] **Step 1.3 — `build/scripts/baseline.js` written**
- [x] **Step 1.4 — grunt build run inside `eo` container (7m 14s, exit 0); baseline re-captured from fresh build → `build/scripts/baseline.json`; diff against itself: PASSED**
- [x] **Step 1.5 — smoke test checklist written** (`build/scripts/smoke-test.md`)
- [x] **Step 1.6 — CI gate added** — Node 16→20, `THEME=euro-office`, `--skip-imagemin`, baseline diff step. Commit `8a0d2bfd50`.
- [x] **Step 4.0 — IE removal complete** — 12 ie/ bundles gone, fix-ie-compat.js deleted, baseline re-captured (20 bundles, 16,706 files, self-diff PASSED). Commit `5cc010e0d7`.
- [x] **Step 4.1 — Bower deleted** — `build/bower.json`, `build/.bowerrc`. Commit `ecc94005ce`.
- [x] **Step 4.2 — chcp removed** — Windows codepage detection block dead on Linux. Commit `4dbebe57b5`.
- [ ] **Step 4.3 — imagemin replacement** — deferred; imagemin never runs (`--skip-imagemin` everywhere); needs sharp/svgo spike.
- [ ] **Step 4.4 — grunt-text-replace** — no CVEs; low urgency; working.
- [x] **Step 4.5 — Makefile branding** — DROPPED. Ascensio attribution in copyright banner is correct (AGPL); theme `config.json` already sets `publisher_name`/`publisher_url` for branding tokens.
- [x] **Step 4.5 — DROPPED** (Ascensio attribution correct; 4.5 was a no-op)
- [x] **Step 4.6 — DROPPED** (phantom — no web-apps sed exists; server's sed is unrelated)
- [x] **Step 4.7 — multi-repo deps doc + preflight assertion** — `docs/multi-repo-deps.md` + sdkjs `test -d` preflight in Gruntfile. Commit `dc7c2791a5`.

.gitignore
```diff
@@ -42,3 +42,13 @@ selenium-debug.log
 
 # generated at build time by theme.config.mjs
 apps/common/mobile/resources/less/_theme-mobile-overrides.less
+apps/common/main/resources/less/_theme-main.less
+
+# Claude session notes — local only, not for the repo
```
juliusknorr: Would prefer to generally exclude .claude for now and only force add if needed

CLAUDE.md
```diff
@@ -0,0 +1,53 @@
...
+| [`webpack5-bare-common-global-contract.md`](.claude/findings/webpack5-bare-common-global-contract.md) | 58 files use bare `Common.*` — load-order invariant |
+| [`deploy-html-inline-ordering.md`](.claude/findings/deploy-html-inline-ordering.md) | `deploy-html.js` must be followed immediately by `inline-svgs.js` — running deploy-html alone leaves `?__inline=true` script tags as unreachable filesystem URLs → "Not supported version" / blank page on all editors |
+| [`deploy-common-bugs.md`](.claude/findings/deploy-common-bugs.md) | Three bugs found in deploy-common.js during overwrite test: (1) `PRODUCT_VERSION` env var ignored → api.js reports `4.3.0`, eurooffice rejects `< 6`; (2) `@@SRC_ROOT@@` not replaced in `apps/common/*.html`; (3) `inline-svgs.js` must cover `apps/common/` — all fixed |
+| [`phase-e-gap3-deploy-resources.md`](.claude/findings/phase-e-gap3-deploy-resources.md) | grunt `deploy-app-main` copies `main/resources/{help,symboltable,watermark,numbering,img}` per editor — replaced by `deploy-resources.js` |
+
+## Parked / known issues (not blocking)
+
+| Issue | Notes |
```
juliusknorr: Those should go into the issue tracker and if needed be referenced in Claude.md but not sure they are needed in such a wide scope

CLAUDE.md
```diff
@@ -0,0 +1,53 @@
...
+| `themes_thumbnail@2x.png` 404 | `sdkjs/common/Images/…` — outside web-apps scope, sdkjs issue. |
+| Transitions.js icon panel blank | All `btn-transition-*` icons invisible — CSS-class template not updated for SVG migration. See `.claude/icon-migration.md`. |
+| `FormsTab.getView()` throws when opening PDF | `FormsTab.setConfig` (which sets `this.view`) is only called when `config.isFormCreator`. `Toolbar.onDocumentReady` calls `FormsTab.getView()` whenever `isPDFForm`. When user opens a plain PDF (not form-creator), `this.view` is undefined, `getView()` throws instead of returning `undefined`. SDK global error handler catches it. Pre-existing upstream bug — not caused by webpack migration. Fix: guard `_viewsCache` in base `getView`, or move FormsTab view creation to `onLaunch`. |
+
+## Hard rules (do not deviate)
+
+- Do not switch Node versions locally — breaks Claude and other tools
+- Do not commit planning/analysis docs — only code and `.claude/` notes
+- Do not use `noParse` for `locale.js` — superseded by string-replace-loader (see findings)
+- Use `claude-opus-4-8` for advisor reviews — Fable model is not available
```
juliusknorr: Would consider this a developer specific decision

apps/documenteditor/forms/index.html.deploy
```diff
@@ -259,21 +259,30 @@
 
           var require = {
               waitSeconds: 30,
+              baseUrl: window.customBaseUrl || "../../",
+              paths: {
+                  app:           "documenteditor/forms/app",
+                  sdk:           "../../sdkjs/word/sdk-all-min",
+                  jquery:        "../vendor/jquery/jquery.min",
```
juliusknorr, line 266: Unrelated but we should file an issue to get rid of the vendored dependencies in https://github.com/Euro-Office/web-apps/tree/main/vendor and use proper dependency management, so dependabot can suggest updates for it.

CLAUDE.md
```diff
@@ -0,0 +1,53 @@
...
+| [`webpack5-backbone-underscore-safe.md`](.claude/findings/webpack5-backbone-underscore-safe.md) | Backbone/Underscore are SAFE — do not add fixes for them |
+| [`webpack5-bare-common-global-contract.md`](.claude/findings/webpack5-bare-common-global-contract.md) | 58 files use bare `Common.*` — load-order invariant |
+| [`deploy-html-inline-ordering.md`](.claude/findings/deploy-html-inline-ordering.md) | `deploy-html.js` must be followed immediately by `inline-svgs.js` — running deploy-html alone leaves `?__inline=true` script tags as unreachable filesystem URLs → "Not supported version" / blank page on all editors |
+| [`deploy-common-bugs.md`](.claude/findings/deploy-common-bugs.md) | Three bugs found in deploy-common.js during overwrite test: (1) `PRODUCT_VERSION` env var ignored → api.js reports `4.3.0`, eurooffice rejects `< 6`; (2) `@@SRC_ROOT@@` not replaced in `apps/common/*.html`; (3) `inline-svgs.js` must cover `apps/common/` — all fixed |
+| [`phase-e-gap3-deploy-resources.md`](.claude/findings/phase-e-gap3-deploy-resources.md) | grunt `deploy-app-main` copies `main/resources/{help,symboltable,watermark,numbering,img}` per editor — replaced by `deploy-resources.js` |
```
juliusknorr: Nothing committed here so i would also skip that in the claude file.

juliusknorr (commented):
A few comments inline.

chrip (commented):
**Existing comments addressed (not repeating):**
- `.gitignore` `.claude/` exclusion preference (juliusknorr)
- `CLAUDE.md` parked issues scope (juliusknorr: lines 36, 40, 52)
- Vendored dependencies in `vendor/` (juliusknorr: `apps/documenteditor/forms/index.html.deploy:266`)

---

**Summary:** Solid migration — removing grunt and IE baggage is the right move. Four findings below, one being a real bug in CI (sequential steps lose the claimed speedup), two are fragility issues that will bite later, and one cleanup.

---

**1. CI doesn't use `build-pipeline.js` — claimed speedup won't materialize**
`.github/workflows/build.yml`

The PR body says "CI build time: ~7 min → ~2 min" but the CI workflow runs all steps sequentially, duplicating the pipeline logic in shell. The parallelism benefit only exists in `build-pipeline.js`. Either replace the individual steps with a single `node scripts/build-pipeline.js` step, or use GitHub Actions jobs with `needs:` for parallel execution. As-is, CI won't see the claimed improvement.

---

**2. `inline-svgs.js` treats zero substitutions as a hard error**
`build/scripts/inline-svgs.js` line 118

```js
if (totalSubstitutions === beforeEditor) {
    process.exitCode = 1;
}
```

Any future template change that removes inline tags (e.g., switching to bundled sprites) will silently break the build with a confusing error. Downgrade to a warning, or make it opt-in per editor.

---

**3. Hardcoded `baseUrl: "../../"` in HTML templates**
`apps/documenteditor/main/index.html.deploy` line 416 (and all other editor templates)

```js
baseUrl: window.customBaseUrl || "../../"
```

This assumes a fixed deploy layout. The old code relied on implicit requirejs baseUrl resolution, which tolerated different paths. If the app is served from a different subdirectory and `customBaseUrl` is never set, requireJS will fail to resolve all modules. Consider computing `baseUrl` dynamically from `document.location.href`.

---

**4. Dead vendor artifacts: `fetch` and `es6-promise` still deployed**
`build/scripts/deploy-common.js` `simpleVendors` array

`fix-ie-compat.js` was deleted and IE conditional loading removed from HTML. The `fetch` and `es6-promise` polyfills are no longer loaded by any code path, but `deploy-common.js` still copies them to BUILD_ROOT. Not harmful, but unnecessary deployed bytes.

Assisted-by: OpenCode:Qwen3.6-27B

moodyjmz:
Thanks for all four findings — responses inline.

**Finding 1 (CI steps vs build-pipeline.js):** Valid. Will switch `build.yml` to call `node scripts/build-pipeline.js` directly in the same way `e2e.yml` already does. Coming in the next commit.

**Finding 2 (inline-svgs zero-substitutions as hard error):** Intentional. Zero substitutions on a specific editor means the HTML templates didn't contain the expected inline tags — that produces blank editors (documented in the findings catalog: a missed inline pass is "Not supported version" / blank page, high blast radius). The script already has a precedent for the softer path: `device_scale.js` missing gets a warning and a skipped tag, because that miss is benign. The per-editor hard error exists precisely because this miss is not benign. Making it a warning would silently defeat the gate.

**Finding 3 (hardcoded `baseUrl: "../../"`):** Pre-existing from the grunt templates — the webpack migration didn't change it. Worth noting though: it's not actually hardcoded without an escape — every template uses `window.customBaseUrl || "../../"`, so a deployer serving from a non-standard path already has a runtime override hook. The embed templates use a bare `"../../"` but those are fixed-layout contexts.

**Finding 4 (fetch / es6-promise "dead" bytes):** You're right that they're not loaded by any IE conditional bundle — that's gone. But they're not dead: `apps/common/locale.js:170` does a live `window.fetch` feature-detect on every editor load and dynamically `require()`s both polyfills if the browser lacks native fetch:

```js
if ( !window.fetch ) {
    var polyfills = ['../vendor/fetch/fetch.umd'];
    if ( !window.Promise ) {
        require(['../vendor/es6-promise/es6-promise.auto.min'], ...);
    } else require(polyfills, _requireLang);
} else _requireLang();
```

So removing them from the manifest would break localisation on any browser that hits that path.

That said — you've raised a valid broader point. The browsers that lack native `fetch` are IE 11 (EOL June 2022) and Firefox 31–38 (2014–2015, current FF is 151). The browserslist still declares `Firefox >= 31`. There's no good reason to keep supporting Firefox 31 in 2026. Properly removing these polyfills means: raise the browser floor in the browserslist (suggest `Firefox ESR` so it tracks dynamically), remove the `locale.js` feature-detect guard, then drop the manifest entries. That's a product-level browser support decision though, not a build system change, so it's out of scope for this PR.

Filed as a follow-up: #129

chrip (commented):
✅ Resolved

- CI sequential steps (chrip Finding 1) — fixed in `a4d8a5b0` via `build-pipeline.js`
- `.claude/` and `CLAUDE.md` added to `.gitignore`, file removed from commit (juliusknorr)

---

### 🔴 Blocks merge

**1. Missing `Assisted-by:` trailers — all 51 commits**

Every commit has `Co-Authored-By: Claude Sonnet 4.6` but none have the `Assisted-by:` trailer required by AGENTS.md. These are separate conventions; `Co-Authored-By` satisfies GitHub attribution but does not fulfil the policy requirement.

See AGENTS.md: "Add an `Assisted-by: AGENT_NAME:MODEL_VERSION` git trailer to every commit containing AI-assisted content."

Please rebase before merge. A one-liner covers all commits:
git rebase origin/main --exec 'git commit --amend --no-edit --trailer "Assisted-by: ClaudeCode:claude-sonnet-4-6"'



**2. `build/scripts/inline-svgs.js:126` — zero substitutions = hard build exit**

```js
if (totalSubstitutions === beforeEditor) {
    process.exitCode = 1;
}
```
This breaks CI the moment a template legitimately stops using inline SVG tags. Downgrade to console.warn, or add a --strict flag for pipelines that want the hard exit explicitly.

---

### 🟡 Should fix


**3. Hardcoded baseUrl: "../../" in HTML deploy templates (index.html.deploy:416)**


baseUrl: window.customBaseUrl || "../../",
Assumes a fixed deploy-directory depth. Suggest deriving from document.location.href at runtime, or documenting the constraint so operators know customBaseUrl is required in non-standard deployments.

**4. Dead vendor artifacts (build/vendor.manifest.mjs:41–42)**

fetch and es6-promise remain in the manifest after fix-ie-compat.js was deleted and IE paths removed from HTML. ~30 KB of unreachable bytes — please remove both entries.

**5. Stale grunt references (build/README.md:248, 420)**

Both lines still reference grunt --skip-imagemin. Rephrase as past tense or remove the command; the surrounding explanatory context can stay.

**6. mobile:install outside Phase 1 banner (build/build-pipeline.js:257–273)**

mobile:install runs before the Phase 1 banner; its dependents appear inside Phase 1. Output ordering is misleading in CI logs. Move mobile:install inside the phase or add a named pre-phase.

---

### ℹ️ Investigated, not an issue
LaunchController.js window.require — code is window.require({waitSeconds: 0}, …), not a bare call; config is intact; correct.
scroller.less data-uri replacements — all 13 replacements correct; data-uri() is unavailable in webpack's less-loader.
o.kill() try/catch — adequately handles ESRCH from already-exited processes.

Assisted-by: ClaudeCode:claude-sonnet-4-6

moodyjmz:
### ⚠️ Depends on #140 — fold in before/with this migration, or the #89 branding fix regresses

The ONLYOFFICE-branding fix (#139) ships on the **grunt `main`** line. Its source edits (`Main.js` gate, `.deploy` loader logo/title tokens) carry across a merge for free. The build-wiring half does **not**:

- Token substitution for `{{APP_TITLE_TEXT}}` / `{{LOADER_LOGO}}` / `{{LOADER_LOGO_DARK}}` in the editor HTML lives in **`Gruntfile.js` `replace:indexhtml`** on the grunt side.
- This migration deletes `Gruntfile.js`. On the webpack side that logic must live in **`build/scripts/deploy-html.js`**, which today substitutes only `@@SRC_ROOT@@`.

If the migration lands without it: `Gruntfile.js` is gone, the tokenized `.deploy` files survive from #139, `deploy-html.js` doesn't know the new tokens → every editor's loader ships `src="{{LOADER_LOGO}}"` and `<title>{{APP_TITLE_TEXT}} …`. #89 silently un-fixes itself the moment webpack becomes `main`.

**#140** (parked draft, base `build/webpack-migration`) is exactly that `deploy-html.js` change. Merge it into this branch — or cherry-pick its `deploy-html.js` commit — before this PR lands. Flagging here so it doesn't fall through the branch gap.

Aiiaiiio (approved):
Tried it, works well

PR #128, opened 2026-06-18
i18n(nl): complete Dutch translations for all editors

nvdael:
## What
Fills in all missing Dutch (`nl`) locale keys across the document, spreadsheet, presentation, PDF and Visio editors (main + embed/forms), bringing `nl` to parity with the English source.

## Numbers
- **6,398** missing keys added across 10 `nl.json` files
- **2,888 (45%)** auto-filled from existing exact EN->NL pairs elsewhere in the product (consistent, no guessing)
- **3,510** newly translated (formal register)

## Approach / quality
- Preserved HTML tags (`<br>`), placeholders (`%1`, `{0}`) and `&` mnemonics (0 placeholder mismatches)
- Proper nouns and theme/effect/colour names intentionally left unchanged
- Each `nl.json` rebuilt in `en.json` key order; existing values untouched -> clean diff

## Verification
`translation/merge_and_check.py --check` reports `nl: 0` missing keys across all five main editors.

## Note
Translation was assisted by Claude (AI). Manual spot-checks were done by hand on a sample of the strings; a fuller native-speaker review of the longer shortcut/help strings is still recommended.


moodyjmz (approved):
Looks good from a tech pov, needs native review, @nvdael need DCO though https://github.com/apps/dco

nvdael:
> Looks good from a tech pov, needs native review, @nvdael need DCO though https://github.com/apps/dco

Thanks for your review! I am a native Dutch speaker and I Sherry-picked multiple lines, but it is too many for all manual checks.

moodyjmz:
Following up on the native review I flagged as still needed — @nvdael, totally understand why a full manual pass wasn't feasible at this scale, 6,400+ strings is a lot to eyeball by hand. I ran a multi-layered AI-agent review over all ten files — one Dutch-fluent pass per editor — to help close that gap.

Most of the strings read as solid, professional Dutch — the shape/color/function names and the bulk of the shortcut descriptions are in good shape. A handful of spots are worth a second look before merge, grouped below so they're easy to work through.

### A few strings that may not say what they're meant to

- `DE.Controllers.Main.savingText` currently reads as "Sending..." — might be worth comparing against `saveTextText` to confirm it's meant to say "Saving...".
- In `Common.Controllers.Shortcuts`, a handful of `txtLabel*` keys (`txtLabelSave`, `txtLabelIndent`, `txtLabelInsertHyperlink`, etc.) got translated, where the sibling keys nearby (e.g. `EditRedo`) keep the literal identifier untouched — worth double-checking whether those specific keys are meant to stay as-is.
- A cluster of arrow shape names in the PDF editor file (`leftRightArrow`, `upDownArrow`, `downArrowCallout`, and several others) use "peil" where the other arrow shapes nearby use "pijl" — probably just want these aligned to one term.
- `mniSentenceCase` / `mniToggleCase` (text-case commands) read as referring to font/typeface rather than letter case — appears in both the PDF and spreadsheet locale files.
- `PDFE.Views.FormSettings.textRadiobox` — "Radial knop" — might be worth comparing with `tipRadioBox`/`textRadioChoice` a few lines away, which use a different term for the same control.
- `Common.Controllers.Shortcuts.txtDescriptionEquationChangeAlignmentRight` currently matches the text of the Left-alignment variant — worth a quick check that it's describing "right" alignment.

### A few small typos/spelling that repeat identically across multiple files

Since these appear word-for-word in more than one locale file, flagging once covers all occurrences:

- `textPluginsSuccessfullyInstalled` — "Plug-Ins zin. Succesvol geïnstalleerd..." — looks like "zin" may have dropped a letter (zijn).
- `Common.UI.Calendar.textShortWednesday` — currently "wij" — likely meant to be the "wo" abbreviation.
- `Common.Utils.ThemeColor.txtLightOrange` — "Lichtorange" — probably meant to be "Lichtoranje".
- `textFromStorage` renders as "Van Opslag" with a capital O in a few places, while one instance elsewhere uses lowercase — worth picking one casing.

### Formal/informal address (u/uw vs je/jouw)

The files are otherwise consistently formal ("u/uw"), which seems to be the established convention. A small set of shared strings — `Common.Controllers.Plugins.helpUseMacros`, `Common.Translation.tipFileLocked`, `Common.Views.ShortcutsDialog.txtEmpty`, and the `tipLicenseExceeded`/`tipLicenseUsersExceeded` pair — use "je/jouw" instead, and the same pattern shows up across several of the editor files. Might be worth a pass to align these with the formal register used everywhere else.

### Terminology that varies within the same file

A few concepts are translated more than one way in the same editor, which might be worth converging on a single term:

- Presentation editor: "Slide Master" appears as both "Diasbasispagina" and "Diamodel".
- Visio editor: "key descriptor" appears as both "sleuteldescriptor" and "sleutelbeschrijving".
- "Restore to default" uses two different verbs in the same shortcuts dialog.

### One process thought, for whenever it's useful

Given the scale here (six thousand-plus strings across ten files), it might be easier on future review — and on catching repeated strings like the ones above — to land this as a few smaller PRs split by app, rather than one combined PR. Also worth checking whether there's an existing translation pipeline (Crowdin/Weblate/etc.) upstream with a shared glossary — that kind of tooling tends to catch exactly this class of consistency issue automatically, and might be worth routing future locale work through if it's already in place. Not a blocker for this PR, just a thought for the workflow going forward.

Happy to make these fixes myself if that's easier than you circling back to it — entirely your call. Great effort getting `nl` this close to full parity either way. Once these are settled I'm comfortable clearing the native-review gate.

nvdael:
Thanks for the thorough review, @moodyjmz — totally understand your points and appreciate the detailed feedback.

Since you already mapped out all the issues so clearly, feel free to apply the fixes yourself if that’s easiest. No problem from my side.

Let me know if you need anything else from me.

Issue #122, opened 2026-06-17
Change SmartPicker behavior


chrip:
This is a follow up of https://github.com/Euro-Office/web-apps/issues/55

* The "SmartPicker" button in the ribbon menu should be renamed to "Add from Nextcloud"
* The tooltip should be changed accordingly
* A click on this button should add a "/" slash at the cursors position and trigger the "Add from Nextcloud" menu. Like in the notes app

<img width="665" height="406" alt="Image" src="https://github.com/user-attachments/assets/433e18e9-224b-4241-82f6-40c47bf097c4" />

* If the user selects something, the inserted content replaces the "/" slash.

chrip:
Implemented:
https://github.com/Euro-Office/eurooffice-nextcloud/tree/fix/issue-122-smart-picker-behavior
https://github.com/Euro-Office/web-apps/tree/fix/issue-122-smart-picker-behavior

PR #130, opened 2026-06-19
Change smart picker behavior

chrip:
This introduced calling smart picker by keystroke "/" anywhere in a doc, ppt, or xls.
Details are in https://github.com/Euro-Office/web-apps/issues/122

For testing you need connector app and web-apps on the following branches:
https://github.com/Euro-Office/eurooffice-nextcloud/tree/fix/issue-122-smart-picker-behavior
https://github.com/Euro-Office/web-apps/tree/fix/issue-122-smart-picker-behavior


apps/spreadsheeteditor/main/app/controller/Toolbar.js
```diff
@@ -258,11 +258,38 @@ define([
...
+                me._smartPickerReplace = '';
+                Common.NotificationCenter.trigger('edit:complete');
+            });
+            // Capture phase: while editing a cell the cell editor consumes the
+            // keydown before it bubbles to document, so a bubble-phase listener
+            // never sees mid-edit "/". Capture runs first, so "/" is caught both
+            // on a selected cell and mid-edit (e.g. "asdf /"); _smartPickerReplace
+            // then drives the "/" removal on insert.
+            document.addEventListener('keydown', function(e) {
+                var key = e.key || (e.originalEvent && e.originalEvent.key);
```
MonaAghili: I believe `e.originalEvent` only exists on jquery wrapped events
This is a native listener  and `e` is already the `KeyboardEvent`
This fallback never runs Just `var key = e.key`

apps/spreadsheeteditor/main/app/controller/Toolbar.js
```diff
@@ -1383,12 +1413,47 @@ define([
...
+                } else if (typeof this.api.asc_insertInCell === 'function') {
+                    // Not editing (selected cell): append to the committed cell
+                    // text via the value path (no formula operators, so no "+");
+                    // drop a trailing "/" trigger if present.
+                    var cell = this.api.asc_getCellInfo && this.api.asc_getCellInfo();
+                    var cur = (cell && cell.asc_getText && cell.asc_getText()) || '';
+                    if (cur.slice(-1) === '/') {
+                        cur = cur.slice(0, -1);
+                    }
+                    this.api.asc_insertInCell(cur + data, Asc.c_oAscPopUpSelectorType.None);
```
MonaAghili: The comment says "no formula operators so no `+`", but that's not really the issue here.
As I investigate
`asc_getText()` internally calls `getValueForEdit()` (`WorksheetView.js:13669`) which it returns the same value shown in the formula bar. For formula cells that's the formula itself (for example, `=SUM(A1:A10)`) not the evaluated result

As a result `cur + data` becomes something like:

`=SUM(A1:A10)https://example.com` (the formula source with the =) when you pass that to `asc_insertInCell` the SDK sees a string starting with `=` and tries to parse it as a formula
The url appended after it makes it syntactically invalid → the cell becomes #NAME?. The original formula is gone.

Even for non formula cells the behavior isn't ideal! If the cell contains `42` concatenation produces `42https://...` converting the numeric value into plain text instead of preserving it as a number

**Suggestion**: if the cell is empty (`cur === ''`)  just insert `data` directly.

If the cell isn't empty, it's not obvious what the expected behavior should be. Should we replace the existing value append to it or reject the operation? It would be better to decide that explicitly instead of silently producing incorrect results for non-empty cells.

j-base64:
Review in progress (started earlier but found some things that need additional checks before posting)

j-base64:
Reviewed asap as requested first

**Method:** AI-assisted (single + multi-angle) + manual testing. 
Tested live (chrome) in documenteditor, spreadsheeteditor, and presentationeditor.

**Verdict:** 👀 some lookups are recommended to the Author

>_**Note:** looks like new commits landed on the PR while doing the review. I checked the diff and dont look to adress any of the points mentionned below, flaggin it in case anything else moved underneath that I missed._

I share my report below, beeing new in the codebase, i've left criticality/priority entirely to your judgment below.

<details>
<summary><strong>📄 Full report here (click to expand)</strong></summary>

### What's working well 🙌
The core happy path is solid: typing `/` in an empty spot, picking "Files," and having it cleanly replace the `/` with a link works exactly as issue #122 describes. Confirmed this in documenteditor, spreadsheeteditor, and presentationeditor. The native caret menu, the provider list, the hand-off to Nextcloud's own picker — all of that came through cleanly.

### Things I think are worth a look before merge 👀

**1. 🟠HIGH - The `/` trigger seems to only work once per session.** 🔍 *Reproduced live in spreadsheeteditor and presentationeditor.* Once typing `/` has opened the contextual menu — whether you go on to pick something and get a link, or just abandon it without picking anything (confirmed both ways, the abandon case in spreadsheeteditor) — the next / in a different cell/textbox gets typed as a literal character instead of opening the menu. It comes back to life only after reloading the document. The AI traced the possible root cause to `installTrigger()` in `SmartPicker.js`: the `lastKey` variable it uses to check "did whitespace precede this `/`" is shared across the whole page and never resets on cell/caret change or after a pick. Same shared file is used by documenteditor, so it almost certainly affects that too, though I didn't explicitly re-test the repeat-use case there. The toolbar button stayed unaffected wherever I tried it.

**2. 🟡MEDIUM - Clicking away doesn't actually truly close the caret menu feature, even though it visually looks like it does.** 🔍 *Reproduced live in documenteditor.* Typed `/`, clicked elsewhere in the document — even far from the popup — the menu appeared to disappear, but pressing Enter afterward, even much later, silently reopened Nextcloud's picker using whatever option was left selected, with no visual sign anything was still pending. The AI then ran a separate scripted check to explain why: the menu's container element is never actually removed from the DOM after that click, meaning its own click-outside handler never runs at all. Likely cause: the editor's own core mouse-handling probably intercepts the click before `SmartPicker.js`'s own click-outside listener ever sees it. Not re-tested in the other two editors, but the code is shared.

**3. ⚪LOW - A keystroke in the narrow gap between picking an option and Nextcloud's dialog opening leaves the trigger stranded.** 🔍 *Reproduced live in documenteditor.* Picking an option closes the caret session immediately, before the reply arrives. Any keystroke landing in that gap — even an unrelated one, like a stray `y` — clears the pending state early, so when the real reply lands the cleanup step gets skipped: I saw both the `/` trigger and the extra keystroke survive intact, sitting right in front of the inserted link (eg "/yhttp://localhost:8082/f/11"
 instead of "http://localhost:8082/f/11"). Traced to `onActivity`/`createPending()` in `SmartPicker.js`. Not re-tested in the other two editors, but the code is shared.


**4. ⚪LOW - The menu doesn't reposition on window resize, despite its own code comment saying it should.** 🔍 *Reproduced live in documenteditor.* Opened the caret menu, resized the browser window without typing anything further — the menu stayed anchored at its original position instead of following the resize, matching a doc-comment in `SmartPickerMenu.js` that describes repositioning as expected behavior that the code doesn't actually implement (no resize listener registered anywhere). Minor compared to the others above — recoverable by just retyping `/` — but a real, confirmed gap between the comment and the code.

### Things the AI found reading the code, that I couldn't reproduce through the UI

Want to be upfront that these didn't hold up under live testing, even though the code pattern itself looks worth a defensive fix:

**5. Two different entry points (the toolbar button and the caret menu) could in theory misattribute each other's replies.** The AI's static review flagged that `createPending()` has no way to tell which reply belongs to which request, and the toolbar button never registers with it at all. I tried to test this directly — open the caret picker, then click the toolbar button before finishing — but couldn't: Nextcloud's own dialog is a focus-trapping modal, and clicking the ribbon while it's open doesn't reach the toolbar in any of the three editors.

**6. A cell being edited via `/` doesn't get the same "don't overwrite existing content" guard as its sibling code path.** Spreadsheet-specific, tested in spreadsheeteditor. The AI noticed `insertLink`'s `isCellEdited` branch in the spreadsheet controller doesn't have the non-empty-cell check that the branch right next to it does (added earlier for a similar formula-corruption concern). I tried feeding it a formula-like value (`=1+1`) through the "Any link" picker myself, to see if it would become a live formula, but Nextcloud's own dialog wouldn't even accept non-URL input, so I couldn't get a malicious value that far. Might be worth a defensive check anyway, given how close it sits to code that was already fixed once for this exact class of bug.

**7. Switching the active cell while a request is pending, in theory, could misdirect the result.** Spreadsheet-specific, tested in spreadsheeteditor. This one also came from the AI's static read, not something I'd spotted myself. I tried to test whether clicking a different cell while Nextcloud's picker dialog is open could cause the link to land in the wrong place, but that dialog turns out to be a proper focus-trapping modal — neither clicks nor keyboard reach the sheet underneath while it's open, so I couldn't actually get into this state through normal use.

### Smaller things

These came out of the AI's static read of the code rather than something I verified by hand:

- Deleting the old `btn-nc-assistant` icon and repointing everything to the new `btn-nc-add` icon also seems to have changed the icon on the unrelated "Ask Nextcloud Assistant" context-menu item (a different, untouched feature), in all three editors. Couldn't verify this live since the Assistant app isn't part of my test setup, but it looked clear from the diff.
- The new unit tests (`slashCanTrigger`, `sanitizeIconUrl`, pending-request bookkeeping) don't cover `insertLink` or the trigger/activity logic — which is exactly the code most of the findings above live in. There's already a very recent commit registering the SmartPicker suite in the test runner though, so it looks like this is being actively worked on 🙂
- The shared trigger/session wiring that the header comment in `SmartPicker.js` describes as consolidated still looks copy-pasted across all three editor controllers, with some drift already visible between the three copies (different comment wording, one has an extra helper the others don't).
- The four `mobile/index.html` files (doc/presentation/spreadsheet/visio) changed only in their hashed CSS filenames — looks like a full local build's output got committed alongside the source changes, might be worth double-checking that's intentional.
- Worth asking: the Assistant-insert wiring feels like a separate feature bundled into a PR titled "Change smart picker behavior" — was that intentional, or would it make sense to split out?
- `AssistantInsert`'s retry loop (the separate "Insert into document" Assistant flow, not something I tested live) gives up after 2 seconds and calls `pluginMethod_PasteHtml` anyway — even though its own code comment says that exact call gets silently dropped by a re-entrancy guard — then reports success unconditionally regardless. On a slow machine, the user's generated content could vanish with no error shown.

### Just sharing, not asking for changes

Also from the AI's static read, minor enough that I didn't chase these further:

- `sanitizeIconUrl`'s regex (shared by all three editors) allows a `/\host/path` form through as if it were same-origin, though it doesn't seem to open anything beyond what's already allowed one line above it.
- `txtCellNotEmpty`, in spreadsheeteditor, has a hardcoded English fallback string that duplicates the locale file entry — the only string in the file doing this.
- A handful of sdkjs-internal DOM ids are hardcoded in several places instead of a shared constant.
- `getHolder()` returns a jQuery-wrapped element under an option named `holderEl`, which reads as if it should be a raw DOM node — small trap for whoever wires up a fourth consumer later.
- Every registered provider's icon gets fetched cross-origin the instant `/` is typed, not just the one eventually picked — a passive "someone is composing" signal to any app with a registered provider.

_Reviewed Assisted by Claude Sonnet 5_

</details>

---

Thanks for all the work on this, happy to pair on any of the above if useful 🙌

chrip:
@j-base64 @MonaAghili @moodyjmz — thanks for the thorough review, that was
really useful. I went through every point and pushed fixes for all of them.
The trigger now survives more than one use per document, clicking away really
does end the session, the resize gap is closed, and the toolbar button no
longer competes with the caret menu. A couple of the spreadsheet items I
answered in code comments rather than by changing behaviour — happy to discuss
if you disagree. Branch is rebased on main, so the old inline comments show as
outdated.

Needs the host side to work: Euro-Office/eurooffice-nextcloud#70.

Ready for another look.

j-base64:
Thanks for the rework 🙌,  update reviewed against my original findings + a new live testing.

**Verdict:** 👀 ✅ Mostly fixed, just suggest a quick look at a few new things surfaced live testing + 💡 UX comments in the full report below. Might be worth getting a casual or UX tester to try this feature.

| # | Issue | Status |
|---|---|---|
| 1 | ~~"/" only working once per session~~ | ✅ Fixed |
| 2 | ~~Clicking away didn't really close the session~~ | ✅ Fixed |
| 3 | ~~Stray keystroke in the hand-off gap~~ | ✅ Fixed |
| 5 | ~~Toolbar/caret entry points could misattribute replies~~ | Not reproducible through UI |
| 6 | ~~Spreadsheet empty-cell guard~~ | ✅ Addressed, no regression |
| 7 | ~~Cell-switch-while-pending~~ | Not reproducible through UI |
| Smaller items | ~~icon, tests, naming, etc.~~ | ✅ Addressed |
| 4 | Menu disappears or jumps vertically on resize | ⚠️ Improved, but another look + UX consideration is suggested (see video in full report) |
| 8 **(NEW)** | **The "/" flow inserts a non-clickable URL when there's text before it** |  Sev. 🟠MEDIUM-HIGH |
| 9 **(NEW)** | **Tiny menu mispositioned (spreadsheet, long cell text)** |  Sev. ⚪LOW |
| 10 **(NEW)** | **"Add from Nextcloud" button's `insertLink` fallback ignores cell state** | ⚠️ Pre-existing bug this PR passed through, video attached in full report, not sure it's a thing this PR should solve (your call)  |

<details>
<summary><strong>📄 FULL REPORT, click to open (6 sections and 2 videos 🎞️)</strong></summary>

### ⚠️ 4. ~~The menu doesn't reposition on window resize, despite its own code comment saying it should.~~ improved but ↠ **menu disappears or jumps vertically on resize**

Dragging the window for real, across documenteditor, presentationeditor and spreadsheeteditor, two issues show up:

- **The menu disappears past a certain window width**, in all three editors.
- **Its vertical position also jumps** between two different offsets mid-drag, rather than following smoothly (seen in presentationeditor).

**A thought, not a request:** for a V1, maybe we could just let the menu close on resize rather than adding code to fix the position - from a UX perspective seems like an acceptable outcome, but your call to decide.



https://github.com/user-attachments/assets/508dff9d-67db-4123-96bc-ecdd581c48f7





### ⚠️ 8. New: the "/" flow inserts a non-clickable URL when there's text before it

Type "Hello ", then "/", pick "Any link" or a file from the tiny menu itself (the correct, guarded path, no toolbar involved) - the URL gets inserted in place of "/", but as plain text: not blue, not underlined, not clickable.

Retested with a bare "/" and no preceding text (empty cell): same flow, same code path, but this time the result is a real, clickable hyperlink (blue, underlined). So the bug specifically needs existing text before the trigger - it's not that this path never produces real links.

This is the guarded, correct-by-design "/" path, this PR's own code (the mid-edit insertion path), not inherited from `main`. Working theory, not confirmed at the code level: this branch inserts the URL as plain text via `pluginMethod_InputText`, and sdkjs likely only auto-converts typed text into a real hyperlink when the URL is the entire cell content - appending it after existing text never triggers that auto-linkify. Haven't traced the exact mechanism further, but the observed behavior (works when the cell was empty, doesn't when there's text first) is consistent and repeatable. Video attached.

### ⚠️ 9. New: tiny menu mispositioned for long cell text (spreadsheet)

Typed a long, multi-word phrase into a cell (e.g. "This is a fairly long sentence with several words "), then "/". The tiny menu opened anchored at the cell's left edge, nowhere near the actual "/" character sitting at the end of the long text, far to the right. Confirmed reproducible.

Not investigated at the code level, but the anchor function for the spreadsheet case (`asc_getActiveCellCoord`) reads the active cell's own coordinates rather than tracking a text-cursor position - a phrase short enough to fit wouldn't reveal the gap between "the cell's position" and "the caret's position within the cell," but a long one does. Low severity, as flagged.

### ⚠️ 10. New: the "Add from Nextcloud" button's `insertLink` fallback ignores cell state entirely (potentially not attributable to this PR, your call)

Three clean repros, all via the toolbar button (video attached):

1. **Non-editing cell, already containing text, whole cell selected** → content is replaced entirely by the raw URL, no warning.
2. **Mid-edit, only part of the cell's text selected** → the *entire* original text becomes the hyperlink's display text, not just the selection - no data lost, but the selection is silently ignored.
3. **Mid-edit via the "/" trigger, cell contains just "/"** → the lone "/" becomes the hyperlink's display text.

All three are one bug, not three - the toolbar button never checks what's already in the cell before acting, unlike the "/" path.

This behavior predates this PR - it's not something introduced from scratch here. Worth noting though: this PR did briefly have a safer version of this same button earlier in its own history, before ending up back on the old behavior. So not simply "someone else's problem" either - potentially worth fixing here rather than deferring.

Case 3 above is not the same thing as case 5's concern: only one thing is happening in that repro (the toolbar button acting on a cell that happens to contain "/"), not two overlapping requests.


https://github.com/user-attachments/assets/989164c0-3124-4b38-bc21-2dc1d9cb6f4e





### 💡 A few UX suggestions

- For the resize on issue 4 , maybe we could just let the menu close instead of chasing pixel-perfect re-anchoring  - seems very acceptable for a v1 if solving it properly turns out tricky.
- Triggering off "/" anywhere may not be obvious to an average user, even if it's a familiar pattern from tools like Notion. Might be worth a toggle to disable it for people who find it surprising rather than convenient.
- "Add from Nextcloud" is narrower than what the feature does - "Any link" inserts a link to anything, not something Nextcloud-specific as far as i understand. Maybe Worth revisiting the name.
- Related: the feature has three different names depending on where you encounter it - no title on the floating menu, "Smart Picker" on the backdropped dialog, "Add from Nextcloud" on the toolbar button. Worth converging on one.
- Not sure what "Any link" inside "Add from Nextcloud" or Smart Picker adds over the already-existing "Insert Link" toolbar feature. Reads as a near-duplicate feature to an end user, just flagging it.
- Not specific to this PR: once a cell becomes a link, clicking it navigates the link instead of letting you edit the cell. For an average user, editing or removing that link afterward will be difficult. Worth a mention for a potential new issue.

### ✅ Smaller things from my original review

- Assistant icon: fixed, confirmed live - context menu's "Ask Nextcloud Assistant" shows its own sparkle icon (`btn-nc-assistant`), separate from the toolbar's `btn-big-nc-add`.
- `mobile/index.html` churn: gone, diffs clean against `main`.
- Shared trigger/session wiring drift across the three controllers: addressed - `install()` now consolidates the Gateway wiring that was three drifting copies.
- Test coverage gap: addressed - the new `SmartPicker.js` test file covers `slashCanTrigger`, `sanitizeIconUrl`, `triggerStillThere`, `createPending`, `createPending.activity`, and `installTrigger` itself (including the click-away and repeat-trigger regressions).
- `sanitizeIconUrl`'s `/\host/path` gap: closed, with a test.
- `txtCellNotEmpty` hardcoded English fallback: gone, now sourced from the locale file like the rest of the controller's strings.
- `holderEl`/jQuery-vs-raw-DOM naming trap: addressed (renamed to `holder`, documented as jQuery-wrapped).
- AssistantInsert's retry-then-call-anyway bug: fixed - falls back to `pluginMethod_PasteText` after the retry budget instead of calling the guarded method anyway, only warns when there's nothing left to insert.

</details>

-----

Hope these findings were useful 🙌, thanks for your updates

-----

_Reviewed Assisted by Claude Sonnet 5_

chrip:
Tested live on the current heads of both branches (web-apps 5f93c11b,
connector a972ed90) against Nextcloud 35.0.0 beta 2 / eurooffice 11.0.4,
in all three editors.

Verified fixed by execution, not by reading the diff: finding 1 (repeat
trigger), 2 (click-away), 3 (stray keystroke in the hand-off gap), 4 (resize),
9 (menu anchor with long cell text), 10 (toolbar button cell guard), and the
AssistantInsert retry/fallback — all three branches: PasteHtml, PasteText
fallback after the 2s budget, and the warning when there is nothing to insert.

Findings 5 and 7 are unreachable by design — clicking the ribbon closes the
caret menu first, and the cell editor ignores programmatic navigation while
active. Both guards still behave correctly when the condition is forced.

Finding 8 is not a defect in this PR. A spreadsheet cell never linkifies a URL
that follows text — typing "Hello https://example.org" by hand gives plain text
too, and a partial-cell hyperlink is not representable in xlsx. The same flow in
documenteditor and presentationeditor produces real, clickable links. So the "/"
path is doing exactly what typing does; the empty-cell/non-empty-cell asymmetry
is the format's, not the branch's.

Nothing found that needs a code change.

Assisted by Claude Opus 5

MonaAghili (approved):
LGTM!

chrip:
Some cases I tested:

Pick a profile:
screencast-16-13-59

Pick a file:
screencast-16-14-31

Insert a http link:
screencast-16-15-06

select smartpicker from menu:
screencast-16-15-48

abort and hide smartpicker, resize frame
screencast-16-17-01

"ask assistant" context menu
screencast-16-20-01

in spreadsheets editor
screencast-16-54-24

in presentation editor
screencast-16-56-08

chrip:
I made the icons smaller, to look more like native editor icons:
image

PR #134, opened 2026-06-22
fix: gate document:ready on script:loaded in PDF editor

Alex-Arsys:
The 500ms setTimeout before triggering document:ready was a race condition: postLaunchScripts (including the FormsTab view class) load asynchronously after app:ready, but in dev mode 30+ individual files consistently take longer than 500ms to load. When Toolbar.applyMode called FormsTab.setConfig before PDFE.Views.FormsTab was defined, createView received undefined as the view constructor, leaving _viewsCache in an inconsistent state and triggering EditingError -25 ("An error occurred during the work with the document").

After the 500ms minimum, check LaunchController.isScriptLoaded(). If scripts are still loading, defer fireDocumentReady until script:loaded fires — the same pattern already used throughout DocumentHolderExt, LeftMenu and Toolbar in this editor.

Fixes #38

Assisted-by: ClaudeCode:claude-sonnet-4-6

### DIFF
--- a/apps/pdfeditor/main/app/controller/Main.js
+++ b/apps/pdfeditor/main/app/controller/Main.js
@@ -1135,8 +1135,7 @@ define([
 
                     me.appOptions.isRestrictedEdit && me.api.asc_SetHighlightRequiredFields(true);
 
-                    var timer_sl = setTimeout(function(){
-
+                    var fireDocumentReady = function() {
                         toolbarController.createDelayedElements();
                         toolbarController.activateControls();
                         documentHolderController.applyEditorMode();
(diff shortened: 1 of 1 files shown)


moodyjmz:
> _Reviewed by **Claude (Sonnet 5)**, posted by @moodyjmz. Verified against `origin/main` source._

Fix is correct, but no other editor (`documenteditor`, `presentationeditor`, `visioeditor`) gates `document:ready` on `script:loaded` — this is new behavior, not restored parity. And there's no timeout fallback or error handling anywhere in this codebase for a failed script load (`LaunchController`'s `require()` calls pass no error callback, `waitSeconds: 0` disables RequireJS's own timeout, and there's no `requirejs.onError`/`onScriptError` defined anywhere). If `app_pack.js` fails to load, the editor now hangs indefinitely instead of degrading after the old flat 500ms.

Needs a bounded fallback before merge — race the `once('script:loaded', ...)` listener against a timeout rather than waiting on it unconditionally.

PR #138, opened 2026-06-23
fix(SSE): smooth tab bar scrolling with mouse wheel and buttons

### LINKED ISSUE #121: Spreadsheet sheet tabs scrolling is slow
Scrolling left-right to navigate the tabs is a bit clunky, especially when there are many tabs

Image

The buttons jump unexpected small amount, scrolling with mouse wheel should also work

maxm11:
## Summary

Fixes #121 — scrolling the spreadsheet sheet-tab bar was clunky: the buttons jumped an inconsistent (often tiny) amount and the mouse wheel did not scroll smoothly.

The tab bar (`<ul id="statusbar_bottom">`) is the scroll container, but all navigation routed through `setTabVisible('forward'/'backward')`, which snaps just far enough to reveal the next partially-hidden tab instead of scrolling by pixels. That caused the erratic button jumps, and the mouse wheel (throttled on Mac) fed the same discrete snap.

## Changes

- **Mouse wheel → smooth pixel scrolling.** `_onMouseWheel` now reads the real `wheel` event delta (both `deltaX` trackpad swipes and `deltaY` wheels), normalizes line/page deltas to pixels, and applies it directly to `scrollLeft`. Removed the Mac-only throttled handler and the legacy `mousewheel`/`DOMMouseScroll` branch — a single `wheel` listener covers all browsers.
- **Buttons → consistent page scroll.** New `scrollByPage()` moves ~one visible width per click instead of the unpredictable next-tab snap.
- Both go through one `scrollByDelta` helper that keeps the trailing spacer (so the last tab clears the overlapping status-bar buttons), handles RTL's negative-scroll model, lets the browser clamp `scrollLeft`, and fires `checkInvisible()` to refresh the arrow-button enabled state.
- Wired up in the spreadsheet and visio editors (both share the `TabBar` component). The numeric `setTabVisible(index)` path used to scroll the active sheet into view is untouched.

## Testing

All three files pass `node --check`. Worth a manual smoke test in a built editor across LTR/RTL and trackpad/notched-wheel before merge.

Agentic Model: Claude Opus 4.8

### DIFF
--- a/apps/common/main/lib/component/TabBar.js
+++ b/apps/common/main/lib/component/TabBar.js
@@ -350,9 +350,6 @@ define([
             // a horizontal trackpad gesture reports deltaX, a vertical wheel deltaY -
             // either one navigates the tab bar
             var delta = e.deltaX || e.deltaY;
-            if (delta === undefined || delta === null) { // legacy mousewheel/DOMMouseScroll
-                delta = e.wheelDelta ? -e.wheelDelta : (e.detail || 0);
-            }
             if (!delta) return;
 
             // normalize line/page based deltas to pixels for a consistent feel
--- a/apps/common/main/lib/component/TabBar.js
+++ b/apps/common/main/lib/component/TabBar.js
@@ -295,12 +295,7 @@ define([
                 elem.addEventListener ? elem.addEventListener( type, fn, false ) : elem.attachEvent( "on" + type, fn );
             };
 
-            if (Common.Utils.isMac) {
-                this.$bar[0].addEventListener('wheel', _.bind(this._onMouseWheelThrottled, this));
-            } else {
-                var eventname=(/Firefox/i.test(navigator.userAgent))? 'DOMMouseScroll' : 'mousewheel';
-                addEvent(this.$bar[0], eventname, _.bind(this._onMouseWheel,this));
-            }
+            this.$bar[0].addEventListener('wheel', _.bind(this._onMouseWheel, this), {passive: false});
             addEvent(this.$bar[0], 'dragstart', _.bind(function (event) {
                 event.dataTransfer.effectAllowed = 'copyMove';
             }, this));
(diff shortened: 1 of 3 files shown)


apps/common/main/lib/component/TabBar.js
```diff
@@ -349,48 +344,54 @@ define([
...
-                }
+            // nothing to scroll - let the event bubble through untouched
+            if (this.$bar[0].scrollWidth <= this.$bar[0].clientWidth) return;
+
+            // a horizontal trackpad gesture reports deltaX, a vertical wheel deltaY -
+            // either one navigates the tab bar
+            var delta = e.deltaX || e.deltaY;
+            if (delta === undefined || delta === null) { // legacy mousewheel/DOMMouseScroll
+                delta = e.wheelDelta ? -e.wheelDelta : (e.detail || 0);
             }
```
MonaAghili: `e.deltaX || e.deltaY` on line 352 both of them are always numbers on a wheel event they never will be `undefined` or `null`!!
so the condition on 353 is never true
`e.wheelDelta` and `e.detail` are both undefined on wheel events anyway, so `delta` would become 0 and the `!delta` check on line 356 would return
I think
```
if (delta === undefined || delta === null) { // legacy mousewheel/DOMMouseScroll
    delta = e.wheelDelta ? -e.wheelDelta : (e.detail || 0);
}
```
 is dead code and should be dropped(lines 353–354 entirely)

MonaAghili:
Other than the dead code comment I left inline every other thing LGTM.

apps/common/main/lib/component/TabBar.js
maxm11: Good catch, thanks — you're right that this handler only ever receives `WheelEvent` instances (bound via `addEventListener('wheel', ...)`), so `deltaX`/`deltaY` are always numbers and the legacy `wheelDelta`/`detail` fallback is unreachable. Dropped it in 5772a6a.

maxm11:
@MonaAghili addressed your comment on the mousewheel fallback — replied inline and pushed 5772a6a. Note: I don't have permission to formally request a re-review on this fork-based PR, so flagging here instead.

MonaAghili (commented):
Review: PR #138 — smooth tab-bar scrolling (post 5772a6a)
Scope: diff 3016a15..HEAD, three files — TabBar.js, and the two Statusbar.js call sites. The previous dead-fallback issue is confirmed gone; this is a fresh pass.

1. RTL scroll direction is inverted — likely regression
Severity: High
Location: apps/common/main/lib/component/TabBar.js, scrollByDelta, line 380


this.$bar[0].scrollLeft += this.isRTL ? -delta : delta;
setTabVisible (lines 542–635, unchanged by this PR) navigates RTL sheets by always adding a positive offset to scrollLeft() — it only changes which tab/edge it inspects for RTL, never the sign of the offset it applies. scrollByDelta — now shared by the mouse wheel and the forward/backward buttons via scrollByPage — flips the sign whenever isRTL is true. Two code paths mutate the same scrollLeft on the same element with opposite sign conventions for RTL. Whichever native RTL scrollLeft model the browser uses, setTabVisible already handles it correctly without a flip, so the new negation is very likely wrong.

Impact: For RTL sheets (isRtlSheet/setDirection(true)), wheel scrolling and the forward/backward status-bar buttons will move the tab bar the wrong way, and checkInvisible's first/last state (which does not flip sign) will disagree with what actually happened — buttons could get stuck disabled/enabled incorrectly.

Suggested fix: Drop the isRTL negation in scrollByDelta and rely on the raw wheel/page delta sign, matching setTabVisible's convention. Verify against a real RTL sheet with many tabs before merging — this is the highest-risk surface in the PR.

2. Ctrl+Wheel (zoom) is hijacked by the tab bar
Severity: High
Location: TabBar.js, _onMouseWheel, line 346–363

Pre-PR, the tab-bar wheel handlers never called preventDefault. Now, whenever the bar overflows, any wheel event — including one with e.ctrlKey === true — is prevented and converted into a horizontal scroll. Ctrl+Wheel is the standard zoom gesture (browser-level and/or app-level); a cursor resting over the bottom tab strip while a user tries to zoom will silently get tab scrolling instead.

Suggested fix: if (e.ctrlKey || e.metaKey) return; before the overflow check, so zoom gestures pass through untouched.

3. deltaX || deltaY picks the wrong axis on diagonal/noisy input
Severity: Medium
Location: TabBar.js, _onMouseWheel, line 352

Precision touchpads commonly emit tiny stray deltaX alongside a large intentional deltaY. e.deltaX || e.deltaY takes whichever is non-zero, preferring a possibly-negligible deltaX over a much larger deltaY, producing barely-perceptible scrolling on gestures that should feel responsive — undercutting the PR's own goal.

Suggested fix: pick by magnitude: Math.abs(e.deltaX) > Math.abs(e.deltaY) ? e.deltaX : e.deltaY.

4. Buttons lost tab-boundary snapping
Severity: Medium
Location: apps/spreadsheeteditor/main/app/view/Statusbar.js:784, apps/visioeditor/.../Statusbar.js equivalent, and TabBar.js scrollByPage line 387-390

onBtnTabScroll used to call setTabVisible(action), which scans tabs and snaps scrollLeft so the next hidden tab's edge lands flush with the bar edge. It now calls scrollByPage(action), which just adds/subtracts a fixed barWidth - 40 px regardless of actual tab widths. With mixed long/short sheet names, button clicks will often stop mid-tab, showing a partially clipped tab — a visible polish regression on button-driven navigation (not the wheel behavior this PR targets).

Suggested fix: either keep setTabVisible for the button handlers (only route wheel input through the new pixel-delta path), or have scrollByPage snap to the nearest tab boundary after applying the page offset.

5. Trailing spacer never removed (pre-existing pattern, now duplicated)
Severity: Low
Location: TabBar.js, scrollByDelta, line 374

Same one-way .separator-item append that already existed in setTabVisible; not new, but worth a one-line comment since this PR touches the same logic. If tabs shrink back to fit, the spacer lingers.

6. Dead defensive guard
Severity: Nit
Location: TabBar.js, line 362 — if (e.preventDefault) e.preventDefault();

Leftover from the removed IE attachEvent path; the listener is now always attached via addEventListener, so e.preventDefault is always present. Harmless, just noise.

Things that look solid
Unifying Mac/non-Mac/Firefox-legacy wheel handling into a single 'wheel' listener with {passive: false} is the right simplification and directly addresses the reported jank — no objection to the core approach.
Early bail-out when scrollWidth <= clientWidth correctly lets the event bubble for page/other scrolling when there's nothing to scroll.
deltaMode normalization (LINE/PAGE → pixels) is a reasonable, standard heuristic; the 16px/line constant is a common approximation, not a bug.
Empty/single-tab edge cases are handled (tabs.length === 0 guard, early return when nothing overflows).
No leftover references to the removed _onMouseWheelThrottled/_lastWheelTime.
Tests
No existing tests cover TabBar.js at all. Given this touches scroll direction math, recommend adding (even lightweight/manual-driven) coverage for:

wheel delta → scrollLeft sign, both LTR and RTL
scrollByPage('forward'/'backward') producing checkInvisible() results consistent with setTabVisible
ctrl+wheel is not intercepted when the bar overflows
Overall assessment
The core fix (unified wheel listener, continuous scrollLeft updates instead of the old throttled tab-jumping) is a good, well-targeted solution to issue #121. But scrollByDelta's RTL sign flip (#1) and the missing ctrl-key guard (#2) are concrete, plausible regressions in scope areas the PR is supposed to cover (RTL layouts, zoom gestures) — not hypothetical nitpicks.

Merge readiness: Not yet — request changes on #1 and #2 before merge; #3 and #4 are worth fixing in the same PR but are not release-blocking on their own.

Remaining risks: #1 is reasoned from static analysis of the sign convention mismatch between scrollByDelta and setTabVisible, not from a live RTL browser test — I'd treat it as high-confidence but verify empirically before demanding a rewrite, since the actual browser RTL scrollLeft model has engine-specific history.

Manual testing to run before approval:

RTL sheet (isRtlSheet = true) with 15+ tabs: wheel scroll both directions, and click both nav buttons — confirm the visually-first/last tab and button disabled-state match expectations.
Ctrl+Wheel while hovering the tab strip — confirm zoom still works.
A precision touchpad diagonal swipe over the tab bar (macOS trackpad ideal) — check for sluggish/near-zero scroll response.
Many long sheet names + forward/backward buttons — check for tabs left partially clipped at the bar edge.
Rapid wheel bursts (momentum scroll) — visually confirm no jank/flicker, and that the trailing separator doesn't visibly shift layout when first appended.

PR #146, opened 2026-06-26
mobile: change add button icon on toolbar

n-goncalves:
Simple change, uses the `add-plus` icon available on Android and iOS for the add button that's on the toolbar for all editors.
It makes more sense for me given the context. 

Here's a before/after on Android:   
Before: 
image

After: 
Screenshot_20260626_160112


n-goncalves:
Bump.

To add some more info. 'icon-plus' is also what's currently used in the Onlyoffice Android app and that's what was used when mobile support was removed from the web-apps repo (cf: https://github.com/ONLYOFFICE/web-apps/commit/dc4c20081bf4a48800de61b7b1a18dade24a18ac#diff-5717fdf8cb39a37d3c84d759296800032a35be1f19aa2e887b7e9061fb4e3cb0L17)

Issue #129, opened 2026-06-19
Raise browser floor and remove fetch/es6-promise polyfills


moodyjmz:
## Background

Flagged during review of #109 (build system migration). Currently \`apps/common/locale.js:170\` guards localisation loading behind a \`window.fetch\` feature-detect:

\`\`\`js
if ( !window.fetch ) {
    var polyfills = ['../vendor/fetch/fetch.umd'];
    if ( !window.Promise ) {
        require(['../vendor/es6-promise/es6-promise.auto.min'],
            function () { require(polyfills, _requireLang); });
    } else require(polyfills, _requireLang);
} else _requireLang();
\`\`\`

This keeps \`vendor/fetch/\` and \`vendor/es6-promise/\` as live deployed artifacts. The browsers that actually hit this path are IE 11 (EOL June 2022) and Firefox 31–38 (released 2014–2015). Current Firefox is 151.

The declared \`browserslist\` still includes \`Firefox >= 31\`.

## Immediate task (the fetch/promise polyfills)

This is a **product-level browser support decision**, not a build system change — hence out of scope for #109.

1. Agree on a new browser floor. Suggest \`Firefox ESR\` in the browserslist (resolves dynamically to the current ESR, no manual updates needed). Drop \`Firefox >= 31\`.
2. Remove the \`window.fetch\` / \`window.Promise\` guard block in \`apps/common/locale.js:170–179\`.
3. Remove \`fetch\` and \`es6-promise\` entries from \`build/vendor.manifest.mjs\` — \`deploy-common.js\` stops copying them, \`verify-deploy.mjs\` stops asserting them.
4. Delete \`vendor/fetch/\` and \`vendor/es6-promise/\` directories.

Do steps 2–4 in a single atomic commit after step 1 is agreed, so the removal is explicit and auditable.

Also update \`vendor/framework7-react/package.json\` browserslist (currently also declares \`Firefox >= 31\`).

## Broader opportunity — audit legacy vendor/ for native equivalents

The browser floor raise is a good moment to audit the full \`vendor/\` tree for things that are now covered by native browser APIs. Quick scan of what's there:

| Vendor | Status | Notes |
|--------|--------|-------|
| \`fetch/\` | **Remove** | Native \`fetch\` universal since Chrome 42 / FF 39 / Safari 10.1 |
| \`es6-promise/\` | **Remove** | Native \`Promise\` universal since Chrome 32 / FF 29 / Safari 7.1 |
[...]

PR #153, opened 2026-06-30
Browser-floor cleanup: drop legacy polyfills + modernise wheel/scroll

moodyjmz:
Low-hanging-fruit browser-floor cleanup — removes dead weight, no behavioural risk on the modern floor.

- 🔥 Stop deploying `fetch` / `es6-promise` polyfills (unused once IE/legacy is off the floor).
- 🐛 Replace `DOMMouseScroll` / `mousewheel` UA-sniffing with the standard `wheel` event across desktop editors.

<details>
<summary>Details</summary>

**Polyfills** — `fetch` and `es6-promise` are dropped from `build/vendor.manifest.mjs`, so the deploy step stops copying them. The vendor dirs stay in the tree (upstream keeps shipping them); we just don't deploy them.

**Wheel/scroll** — three fixes in the desktop editors' `Print.js` / `DocumentHolder.js`:
1. Replace the `DOMMouseScroll`/`mousewheel` UA-sniff with the standard `wheel` event.
2. Switch `Print.js` binding from jQuery `.on()` to `addEventListener`.
3. Correct the wheel `delta` sign in `Print.js` and `DocumentHolder.js`.

Part of the browser-floor-stage1 work; the mobile/ES2022 half is a separate PR.
</details>

### DIFF
--- a/apps/documenteditor/main/app/controller/Print.js
+++ b/apps/documenteditor/main/app/controller/Print.js
@@ -131,8 +131,7 @@ define([
             }, this));
             Common.NotificationCenter.on('margins:update', _.bind(this.onUpdateLastCustomMargins, this));
 
-            var eventname = 'wheel';
-            this.printSettings.$previewBox.on(eventname, _.bind(this.onPreviewWheel, this));
+            this.printSettings.$previewBox[0].addEventListener('wheel', _.bind(this.onPreviewWheel, this), {passive: false});
         },
 
         setMode: function (mode) {
--- a/apps/documenteditor/main/app/controller/Print.js
+++ b/apps/documenteditor/main/app/controller/Print.js
@@ -526,7 +526,7 @@ define([
                 e.preventDefault();
                 e.stopImmediatePropagation();
             }
-            var forward = (e.deltaY || (e.detail && -e.detail) || e.wheelDelta) < 0;
+            var forward = e.deltaY > 0;
             this.onChangePreviewPage(forward);
         },
 
--- a/apps/documenteditor/main/app/controller/Print.js
+++ b/apps/documenteditor/main/app/controller/Print.js
@@ -131,7 +131,7 @@ define([
             }, this));
             Common.NotificationCenter.on('margins:update', _.bind(this.onUpdateLastCustomMargins, this));
 
-            var eventname = (/Firefox/i.test(navigator.userAgent))? 'DOMMouseScroll' : 'mousewheel';
+            var eventname = 'wheel';
             this.printSettings.$previewBox.on(eventname, _.bind(this.onPreviewWheel, this));
         },
 
--- a/apps/common/main/lib/component/TabBar.js
+++ b/apps/common/main/lib/component/TabBar.js
@@ -344,10 +344,12 @@ define([
         },
 
         _onMouseWheelThrottled: function(e) {
-            var delta = e.deltaY !== undefined
+            var rawDelta = e.deltaY !== undefined
                 ? -(e.deltaMode === 1 ? e.deltaY * 30 : e.deltaY)
                 : ((e.detail && -e.detail) || e.wheelDelta);
-            if (Math.abs(delta) < 10) {
+
+            this._wheelAccum = (this._wheelAccum || 0) + rawDelta;
+            if (Math.abs(this._wheelAccum) < 10) {
                 return;
             }
 
@@ -356,6 +358,9 @@ define([
                 return;
             }
             this._lastWheelTime = now;
+
+            var delta = this._wheelAccum;
+            this._wheelAccum = 0;
 
             var hidden = this.checkInvisible(true);
 
--- a/apps/common/main/lib/component/TabBar.js
+++ b/apps/common/main/lib/component/TabBar.js
@@ -295,12 +295,7 @@ define([
                 elem.addEventListener ? elem.addEventListener( type, fn, false ) : elem.attachEvent( "on" + type, fn );
             };
 
-            if (Common.Utils.isMac) {
-                this.$bar[0].addEventListener('wheel', _.bind(this._onMouseWheelThrottled, this));
-            } else {
-                var eventname=(/Firefox/i.test(navigator.userAgent))? 'DOMMouseScroll' : 'mousewheel';
-                addEvent(this.$bar[0], eventname, _.bind(this._onMouseWheel,this));
-            }
+            this.$bar[0].addEventListener('wheel', _.bind(this._onMouseWheelThrottled, this));
             addEvent(this.$bar[0], 'dragstart', _.bind(function (event) {
                 event.dataTransfer.effectAllowed = 'copyMove';
             }, this));
@@ -365,7 +360,9 @@ define([
         },
 
         _onMouseWheelThrottled: function(e) {
-            var delta = (e.detail && -e.detail) || e.wheelDelta;
+            var delta = e.deltaY !== undefined
+                ? -(e.deltaMode === 1 ? e.deltaY * 30 : e.deltaY)
+                : ((e.detail && -e.detail) || e.wheelDelta);
             if (Math.abs(delta) < 10) {
                 return;
             }
--- a/apps/common/main/lib/component/TabBar.js
+++ b/apps/common/main/lib/component/TabBar.js
@@ -343,22 +343,6 @@ define([
             return this;
         },
 
-        _onMouseWheel: function(e) {
-            var hidden  = this.checkInvisible(true),
-                forward = ((e.detail && -e.detail) || e.wheelDelta) < 0;
-
-            if (forward) {
-                if (hidden.last) {
-                    this.setTabVisible('forward');
-                }
-            } else {
-                if (hidden.first) {
-                    this.setTabVisible('backward');
-                }
-            }
-            Common.NotificationCenter.trigger('hints:clear');
-        },
-
         _onMouseWheelThrottled: function(e) {
             var delta = e.deltaY !== undefined
                 ? -(e.deltaMode === 1 ? e.deltaY * 30 : e.deltaY)
--- a/apps/documenteditor/main/app/controller/DocumentHolder.js
+++ b/apps/documenteditor/main/app/controller/DocumentHolder.js
@@ -381,10 +381,7 @@ define([
                 me._isScrolling = false;
             }, 100);
 
-            var delta = (_.isUndefined(event.originalEvent)) ? event.wheelDelta : event.originalEvent.wheelDelta;
-            if (_.isUndefined(delta)) {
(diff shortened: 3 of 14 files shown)


apps/documenteditor/main/app/controller/Print.js
```diff
@@ -131,7 +131,7 @@ define([
             }, this));
             Common.NotificationCenter.on('margins:update', _.bind(this.onUpdateLastCustomMargins, this));
 
-            var eventname = (/Firefox/i.test(navigator.userAgent))? 'DOMMouseScroll' : 'mousewheel';
+            var eventname = 'wheel';
             this.printSettings.$previewBox.on(eventname, _.bind(this.onPreviewWheel, this));
```
MonaAghili: as far as I understand `$previewBox` is a jquery obj and in current  version of it which is 3.5.1 the wheel event doesn't expose `deltaY` on the jquery event object which means `e.deltaY` would be `undefined ` so forward wil lbe always false

would it make sense to change the event back to mousewheel in all printjs files?  The jquery.mousewheel plugin can provide `e.deltaY` for `mousewheel` events

moodyjmz: Good catch — you're right. jQuery 3 wraps events and doesn't copy `deltaY` into the wrapper, so `e.deltaY` was `undefined` and `forward` always evaluated to `false`.

Fixed in 230a052: switched all five `Print.js` files from `$previewBox.on('wheel', ...)` to native `$previewBox[0].addEventListener('wheel', ..., {passive: false})`. The `{passive: false}` is needed because `onPreviewWheel` calls `preventDefault()` on Ctrl+scroll — consistent with how DocumentHolder.js already binds its wheel listener.

moodyjmz: Fair point, and you were right about the version you saw — jQuery wasn't exposing `deltaY` on the `wheel` event. Since switched to native `addEventListener('wheel', …)` which does. Apologies for the drive-by review request though — Claude got a bit eager and pinged reviewers when this should've gone up as a draft first. Marking it draft now while I finish the runtime testing. :P

moodyjmz:
**Independent review caught two sign-inversion bugs introduced by this PR — both now fixed in 0ffa044.**

---

### Root cause

The old jQuery `.on('mousewheel', handler)` path routed through `vendor/perfect-scrollbar/src/jquery.mousewheel.js`, which negates native `deltaY` to up-positive (scroll up = positive, matching legacy `wheelDelta`). Switching to native `addEventListener('wheel')` bypasses that plugin entirely. `TabBar.js._onMouseWheelThrottled` was already written to account for this (`-deltaY`). The other two handler families were not.

### Bug 1 — Print preview scroll direction inverted (all browsers)

`onPreviewWheel` in all five Print.js files:
```js
// before: down-scroll → deltaY > 0 → forward = false → previous page (wrong)
var forward = (e.deltaY || ...) < 0;
// after
var forward = e.deltaY > 0;
```
This regressed Chrome, which worked before the PR.

### Bug 2 — Ctrl+scroll zoom direction inverted (Firefox)

`handleDocumentWheel` / `onDocumentWheel` in all five editors:
```js
// before: Firefox has no wheelDelta → fell through to raw deltaY → down-positive
//         scroll up → deltaY < 0 → delta < 0 → zoomOut (wrong)
var delta = (_.isUndefined(event.originalEvent)) ? event.wheelDelta : event.originalEvent.wheelDelta;
if (_.isUndefined(delta)) { delta = event.deltaY; }

// after: negate deltaY → up-positive → existing delta>0=zoomIn stays correct
var delta = event.deltaY !== undefined ? -event.deltaY : event.wheelDelta;
```
Firefox had no working Ctrl+zoom before this PR (DOMMouseScroll path was dead). The fix makes it correct rather than merely swapping dead for inverted.

apps/documenteditor/main/app/controller/Print.js
MonaAghili: @moodyjmz 
No need to apologize!!!  we're a team and the whole point of reviews is to catch things together and make the project better
It happens to every single of us especially in a big codebase  even claude misses things sometimes!

other than that I didn't find any other issues and LGTM!

moodyjmz:
@MonaAghili heads up — rewrote history on this branch to fix a commit-message bug (the first commit said `Closes #129`, which would've auto-closed a tracking issue that's only ~25% done; changed to `Refs #129`).

Content is byte-for-byte identical (trees match exactly, only that one commit message changed), but SHAs shifted:

| old | new |
|---|---|
| `affc120b` | `60c86976` |
| `b602c07c` | `0f14de06` |
| `230a0528` | `8ecd9398` |
| `0ffa044d` | `b71d05ee` |
| `4251ee6a` | `56b5c91b` |
| `8d140ceb` | `b983e3ee` |

Your review comment on `230a0528` is now anchored to a SHA that's no longer on the branch — the thread should still be visible here, just flagging in case it looks odd on your end.

MonaAghili:
Reviewed `fix/browser-floor-stage1` (head `b983e3ee`) fresh — ignoring prior threads, verifying against the current tree rather than the old SHAs.

Re the history rewrite: pulled all six old commits by SHA via the API (still fetchable even though force-pushed off the branch) and diffed each old→new tree pair directly — all six match exactly. Content is genuinely byte-for-byte identical; only `affc120b`'s message changed, from `Closes #129 (fetch/es6-promise portion)` to `Refs #129 (... browserslist floor, locale.js guard removal, and vendor dir deletion are still outstanding)`. Confirmed, not just taken on faith.

## Non-blocking

**`locale.js`'s `window.fetch` guard is now unreachable-but-present — matches #129, worth closing out in the same PR.**

`build/vendor.manifest.mjs` drops `fetch`/`es6-promise` here, so `apps/common/locale.js:170-179`'s `!window.fetch` branch now `require()`s files that no longer exist in a deployed build if it's ever hit. I initially read this as a live risk, but per [moodyjmz's later comment on #129](https://github.com/Euro-Office/web-apps/issues/129#issuecomment-4840818099) — and independently confirmed by grepping — `apps/common/main/lib/util/utils.js:1051`, `Desktop.js:334/338/349/356`, and `Plugins.js:1048/1123` all call `fetch()` unconditionally already, with no feature-detect. So the guard was already dead protection before this PR; nothing here makes it more reachable. Also checked `build/scripts/verify-deploy.mjs` on both `main` and this branch — no `fetch`/`es6-promise` assertions on either, so that part of #129's plan is moot too.

The actual gap: #129's revised scope (the comment, not just the issue body) calls for a build-time string-replace-loader rule to strip the dead guard block out of `locale.js` at build time, specifically so the fork stays merge-clean against upstream (upstream still ships/uses this guard). This PR does the manifest half of that atomic scope without the loader rule. Not a correctness risk, but worth folding in here or opening the follow-up now rather than leaving `#129` partially done with no tracking pointer in this PR's description.

**`TabBar.js` — throttle/threshold now applies to platforms that didn't have it before.**

Non-Mac platforms used to scroll the tab bar via the untouched `_onMouseWheel` (immediate, no throttle/threshold). This PR routes all platforms through `_onMouseWheelThrottled`, which adds a 50ms throttle and an `Math.abs(delta) < 10` per-event magnitude filter, with no accumulation across ticks. A touchpad emitting small deltaY per tick during inertial scroll (commonly 1-6px on Windows/Chrome) could now be filtered out on every tick, where a physical mouse wheel (~100+ per notch) still passes. Worth a manual touchpad check on Windows/Linux before merge — this behavior change isn't mentioned in the PR description.

**No regression tests for the two direction fixes.**

The print-preview page-nav direction (`onPreviewWheel`) and the ctrl+wheel zoom direction (`DocumentHolder.js`) were both silently inverted for an unknown period and are fixed here — verified correct by tracing through `onChangePreviewPage`/`zoomIn`/`zoomOut`. A cheap synthetic-`WheelEvent` test asserting direction would catch the next accidental sign flip; nice to have, not blocking.

**Duplicated delta-normalization logic across 5 editors.**

`event.deltaY !== undefined ? -event.deltaY : event.wheelDelta` (or its `deltaY > 0` cousin) is repeated verbatim in documenteditor/pdfeditor/presentationeditor/spreadsheeteditor/visioeditor. Same bug, fixed five times. Consider a shared `Common.Utils` helper next time one of these files is touched, so future refinements don't need to be repeated across all five.

## Verified as correct, no issues

- `onPreviewWheel` direction fix (all 5 `Print.js`) — traced through `onChangePreviewPage`, confirms old code advanced pages on scroll-*up*, which was backwards; new code is correct.
- Ctrl+wheel zoom direction fix (all 5 `DocumentHolder.js`) — confirmed old `deltaY` fallback was unflipped and zoomed the wrong way; new code matches legacy `wheelDelta` convention.
- `TabBar.js` Mac wheel bug — before this PR, `_onMouseWheelThrottled` read `e.detail`/`e.wheelDelta` on a native `wheel` event, which never sets those, so `delta` was always `NaN` and tab-bar scroll on Mac could only ever go `'backward'`. Genuinely broken before, fixed now.
- Listener migration (`mousewheel`/`DOMMouseScroll` → `wheel`, jQuery `.on()` → `addEventListener`) — no dangling references to old event names, no new leak (none of these bindings had teardown before or after), `{passive: false}` correctly preserved everywhere a handler conditionally calls `preventDefault()`.
- Dead code removal (`TabBar.js` `_onMouseWheel`) — confirmed unreferenced elsewhere.
- Pre-existing view+document dual wheel binding in `DocumentHolderExt.js`/`DocumentHolder.js` — predates this PR, still benign (`stopPropagation()` on the ctrl branch prevents double-firing).

moodyjmz:
@MonaAghili could you take another look and approve when you get a chance? No rush.

Two commits landed after your last look and haven't had any human eyes on them yet:
- `56b5c91` — drops the now-dead `_onMouseWheel` handler in `TabBar.js`
- `b983e3e` — removes the stale `fetch`/`es6-promise` entries from `build/common.json`

Both are small and I've checked them against the rest of the codebase, but a second pair of eyes on those two specifically would be good. Everything else is unchanged from what you already reviewed (the SHAs just shifted — see my comment above).

Gracias! 🙏

moodyjmz:
@MonaAghili Ah, you did take another look, cool

moodyjmz:
@MonaAghili branch moved again, sorry — two more changes since your last look:

- Reworded the first commit's message (was `Refs #129`, wording tightened, no code change)
- New commit: `TabBar.js` `_onMouseWheelThrottled` now accumulates sub-threshold wheel deltas instead of dropping them, per your touchpad-scroll note — the small per-tick deltas from inertial touchpad scroll no longer get silently swallowed

Current head is `4a5283d`. Could you take another look and approve when you get a chance?

MonaAghili (approved):
LGTM

PR #155, opened 2026-06-30
fix(mobile): show copy/cut/paste icons in context actions

n-goncalves:
On the mobile editor, you can have access to contextual actions based on the selected element. For items that should be shown as icons, those icons aren't imported in the JSX code. That cause the icons to be missing in all editors. 

<details><summary>Show screenshot of the missing icon in spreadsheeteditor</summary>
image
</details> 

This PR fixes it. For now it is only testable in the fallback path since ContextAction is commented out, but it's simple enough to be merged as is. 

Not providing a picture of it working due to https://github.com/Euro-Office/DocumentServer/issues/258 but I have tested it.


moodyjmz:
Thanks @n-goncalves — code looks good. Built all three mobile editors merged into `main` (compiles clean) and verified the icon assets land in the bundle. Still want to give it a run-through with the Mk1 eyeball before merge.

moodyjmz (approved):
Nice work once again @n-goncalves, thanks!

Issue #154, opened 2026-06-30
CI: add mobile-editor validation to build gates and e2e (regression #258)


moodyjmz:
Follows the regression reported in Euro-Office/DocumentServer#258 (https://github.com/Euro-Office/DocumentServer/issues/258).

## Why

#258 shipped a broken mobile editor to `latest-dev`: the mobile build output
(`index.html`, `dist/`, `css/`, `locale/`) never reached `BUILD_ROOT`, leaving a
`resources/`-only folder and a 404. It reached a published image because nothing in CI
tests mobile:

- `build/scripts/verify-deploy.mjs` asserts vendor / embed / main / forms artifacts —
  **no mobile assertions**.
- `.github/workflows/e2e.yml` sets `SKIP_MOBILE: '1'` and drives desktop editors only —
  mobile is never built or loaded.

Two blind spots, one regression. We want the equivalent of the existing smoke/manifest
gates, extended to mobile.

## Proposed — Tier 1: build-time manifest gate (fast, runs in every build incl. the bake image)

Extend `verify-deploy.mjs` to assert, per mobile editor (documenteditor,
spreadsheeteditor, presentationeditor, visioeditor):

- `apps/<editor>/mobile/index.html` and `index_loader.html` exist and are non-empty
- `apps/<editor>/mobile/{dist,css,locale}/` exist and are non-empty

Skip when `SKIP_MOBILE=1`. This alone would have failed the #258 build loudly. Cheap,
no extra infra.

## Proposed — Tier 2: e2e mobile smoke (deep, real browser)

- Drop `SKIP_MOBILE: '1'` in `e2e.yml` (or run a second matrix leg with mobile enabled).
- Add a Playwright test that opens a mobile editor URL (mobile UA / viewport) and asserts
  the app shell mounts — not a 404 / "Not supported version" page.
- Cost: ~50s extra build + one editor load. Acceptable for catching a whole-platform outage.

## Acceptance

- A missing mobile artifact fails the build (Tier 1).
- A mobile editor that loads to a blank/404 page fails e2e (Tier 2).
- Both run on PRs touching `web-apps`.

Depends on / follows the #258 deploy fix (`deploy-mobile.js`).

moodyjmz:
## Status update

### Tier 1 (build-time manifest gate) — ✅ done

Landed in #156 (mobile assertions added to `verify-deploy.mjs`), then made **unconditional** in #157 / #158 (merged).

[...]

PR #156, opened 2026-06-30
Mobile editors: fix not-loading + ES2022 target / i18next crash

### LINKED ISSUE #258: Regression: mobile editor doesn't load in latest-dev
I decided to post it here for visibility even if I assume that the fix will be in `web-apps`. Feel free to move that issue if that's better. .

I am using the docker setup in `develop`. After pulling `ghcr.io/euro-office/documentserver:latest-dev` this morning, the mobile editor stopped working. 
The root cause is that the `mobile` folder where the app is normally stored only contains the resources. So when it tries to load the `index.html`, you get a 404 error.  See the attached screenshot below for what it looks like:

Given that the build process was changed yesterday, I'm tagging you @moodyjmz , sorry if that's not your fault. 

```bash
docker images --digests ghcr.io/euro-office/documentserver
REPOSITORY                           TAG          DIGEST                                                                    IMAGE ID       CREATED        SIZE
ghcr.io/euro-office/documentserver   latest-dev   sha256:5c49561919918aa790082c051ab0482fbd75e7e7ec96717a22b43219cb393fd1   c0d61f49a7d2   18 hours ago   7.78GB

docker compose exec eo bash
root@d5ea4ff17091:/# ls /var/www/euro-office/documentserver/web-apps/apps/documenteditor/mobile/
resources
```

For comparison, here are the expected content of that folder using `ghcr.io/euro-office/documentserver:v9.3.2`, (sha256:5a1a8b7083856b309f09a5b06aa96f80182cbdfccddcae6c082ecf7a2b3cbf2b):
```bash
root@f12d8fedf58e:/# ls /var/www/euro-office/documentserver/web-apps/apps/documenteditor/mobile/
css  dist  index.html  index.html.gz  index_loader.html  index_loader.html.gz  locale  resources
```

Image

moodyjmz:
Two mobile fixes that landed together.

1. **Mobile editors don't load** — reported in Euro-Office/DocumentServer#258. The grunt→pipeline migration dropped the mobile deploy step, so the mobile folder shipped with only `resources/` (404). Ported as `deploy-mobile.js` + verify-deploy assertions.
2. **ES2022 target / i18next crash** — the mobile bundle built at esbuild `target:'es2015'` throws `TypeError: we is not a function` on launch. Fixed by centralising the browser target to ES2022.

<details>
<summary>Root cause + validation</summary>

**Crash cause.** The esbuild minifier swap (`f0e183749c`) set `target:'es2015'`, which *downlevels* the whole bundle — including i18next 25's class-based internals. esbuild's ES2015 class-field lowering mis-compiles a Translator internal → `we is not a function`. Not webpack-vs-grunt and not the i18next dep bump (it ran fine under grunt's Terser build) — mobile was always a webpack build; only the minifier changed.

**Fix.** Collapse the four scattered, contradictory browser targets (babel `>0.25%`, esbuild `es2015`, postcss Chrome 49, decided floor) into a single source of truth `build/browser-floor.mjs` at **ES2022** (Nextcloud floor: iOS 17 / Safari 17; Android 9 = updatable WebView). ES2022 keeps class fields native, so the breaking transform never runs. Also: restore `drop_console` (the esbuild swap silently dropped Terser's), make build mode explicit + echoed in the pipeline banner, decouple mobile `watch` to `WATCH=1`, and add a `verify-browser-target` gate that fails the build if a target *literal* (string/array/object) is hardcoded in a config instead of imported from `browser-floor.mjs`. (It's a static-source check — it asserts the import + absence of a literal, not that the imported value is the one wired in.)

**Validation** (production build in the `eo` dev container):
- `__publicField` (esbuild class-field lowering helper) = **0**; 79 native `#` private fields retained → the es2015 transform is gone.
- `i18next is ready` = **0** in all four mobile bundles → `drop_console` restored.
- All four mobile editors load and render in-browser (confirmed manually).
- `verify-browser-target` passes, and fails on a planted hardcoded target (string, `esnext`, and array-literal shapes).

**Review follow-ups (addressed):** the two verify gates were hardened after review —
`verify-browser-target` now rejects array/object/`esnext` target literals (not just
ES-year strings), and `verify-deploy` asserts the actual `dist/js/app.js` entry rather
than only the directory. The build/README pipeline reference was also brought up to the
current 5-phase structure.

</details>

Fixes Euro-Office/DocumentServer#258
Follow-ups: Euro-Office/web-apps#154 (CI mobile validation), Euro-Office/DocumentServer#259 (Makefile dev/prod targets)

---
_Commits are AI-co-authored (Claude Sonnet 4.6) and adversarially reviewed/verified (Claude Opus); the fix was validated against a real production build + browser load. Human diff review still wanted._


MonaAghili:
Hey @moodyjmz reviewed it with help of Claude and I share  it's findings with you

Reviewed the branch diff directly against the grunt configs it replaces and ran the browserslist query live. Two core fixes are correct. One real bug. One new behaviour worth flagging.

---

**Core fix 1: deploy-mobile.js — correct**

Cross-checked every copy step against `build/documenteditor.json` (the grunt config this replaces):

| What gets copied | grunt | deploy-mobile.js |
|---|---|---|
| `index.html` → `index.html` + `index_loader.html` | `copy.index-page` — both map to same source | `copyFileSync` twice from same source ✓ |
| `css/*.css` (no maps) | `copy.assets src: '*.css'` | `copyDirFiltered({ include: ['*.css'] })` ✓ |
| `dist/**` | `copy.assets src: '**'` | `copyDir(distSrc, dest/dist)` ✓ |
| `locale/*` | `copy.localization src: '*'` | `copyDir(localeSrc, dest/locale)` ✓ |
| `resources/img/**` (per-editor) | `copy.images-app` | `copyDir(imgSrc, dest/resources/img)` ✓ |
| `apps/common/mobile/resources/img/**` | `copy.images-common[0]` | `copyDir(commonMobileImg, ...)` ✓ |
| `apps/common/main/resources/img/about/**` | `copy.images-common[1]` | `copyDir(aboutSrc, .../about)` ✓ |

`resources/css/` not copied — that directory doesn't exist in the source tree; grunt's `mobile.css` task was a no-op. Not a gap.

The `.gz` files in the v9.3.2 image (`index.html.gz` etc.) come from `documentserver-static-gzip.sh` in `document-server-package`, which gzips everything in `web-apps/` during packaging — not the web-apps build's job. Not a gap here either.

Phase ordering is correct: `deploy-mobile` (Phase 2) runs strictly after webpack writes to the source tree (Phase 1) and strictly before `deploy-theme-images` (Phase 3), which overwrites the same `resources/img/` paths and must win.

---

**Core fix 2: ES2022 target / i18next crash — correct**

`target: 'es2015'` made esbuild emit `__publicField()` helpers to downlevel class fields. i18next 25's `Translator` uses native class fields. `__publicField` corrupts the prototype chain → `TypeError: we is not a function`. Moving to `['safari17', 'chrome111']` (ES2022) keeps class fields native. Right fix, right signal to check (`__publicField` count = 0).

`dist/js/app.js` stable-name assertion in `verify-deploy` is correct — webpack config confirms `filename: 'dist/js/[name].js'` with no hash, entry name `app`.

---

**Bug: `and_chr >= 111` doesn't work as a version floor**

`build/browser-floor.mjs`:
```js
export const BROWSERSLIST = ['iOS >= 17', 'Safari >= 17', 'and_chr >= 111', 'not dead'];
```

Ran this live against the installed browserslist:
```
and_chr >= 111  →  ['and_chr 132']   // single current version only
chrome >= 111   →  23 entries        // actual floor
```

browserslist does not track historical `and_chr` versions. `and_chr >= 111` resolves to whatever the current `and_chr` is — not "Chrome 111 and above as a floor." The correct query for Android Chrome version floors is `chrome >= 111` (browserslist resolves desktop Chrome historically, and it's the right V8 proxy for Android Chrome).

The `verify-browser-target` gate only checks that `and_chr` appears in the array — it doesn't validate that the version constraint resolves. So the gate passes and this is invisible.

Functionally harmless today (current Chrome >> 111, both support ES2022 natively, so babel/postcss output is identical). But the intent isn't landing.

**Fix:** `'and_chr >= 111'` → `'chrome >= 111'`

---

**New behaviour: `drop: ['console', 'debugger']`**

```diff
-new EsbuildPlugin({ target: 'es2015', css: true })
+new EsbuildPlugin({ target: ESBUILD_TARGET, css: true, drop: env === 'production' ? ['console', 'debugger'] : [] })
```

This was not in the webpack config on `main` (verified). esbuild's `drop: ['console']` removes the entire call expression including arguments — `console.log(fn())` silently drops `fn()` in production. Intentional for a production bundle, but it's a new behaviour in this pipeline, not a restore.

---

**Verdict: approve. Fix `and_chr >= 111` → `chrome >= 111` inline or as immediate follow-up.**

moodyjmz:
Thanks @MonaAghili — the `and_chr >= 111` catch was right and is fixed in 700138d298 (`chrome >= 111`, which resolves as a real historical floor; `and_chr` is latest-only). Everything else corroborates too.

One point I'd push back on: the `drop: ['console', 'debugger']` "new behaviour, not a restore" framing measures against the wrong baseline. `main` is already post-regression — the esbuild swap is what *removed* the console-dropping. The config immediately before that swap had it under Terser, so re-adding it restores prior pipeline behaviour rather than introducing new behaviour.

The side-effect caveat (`console.log(fn())` drops `fn()`) is real but not new — Terser's `drop_console` did the same thing, so this reintroduces known behaviour, not a novel risk. Production-only, so no dev impact.

<details>
<summary>Evidence</summary>

`f0e183749c^:vendor/framework7-react/build/webpack.config.js:88` (config just before the esbuild swap):

```js
new TerserPlugin({ terserOptions: { compress: { drop_console: env === 'production' } } })
```

The swap replaced Terser with `EsbuildPlugin({ target: 'es2015', css: true })` (now on `main`, no drop). This PR re-adds `drop: ['console','debugger']`.

<sub>Comment prepared with Claude (Opus), verified against the branch.</sub>
</details>

MonaAghili (approved):

MonaAghili (approved):

MonaAghili:
LGTM!

PR #158, opened 2026-07-01
Remove SKIP_MOBILE — always build & gate mobile in CI

### LINKED ISSUE #157: Remove SKIP_MOBILE — always build & gate mobile in CI
Drop `SKIP_MOBILE` so the pipeline always builds and gates mobile. CI skipping mobile (`e2e.yml: SKIP_MOBILE: '1'`) is what let Euro-Office/DocumentServer#258 and the es2015 i18next crash ship with green CI. Web-apps half of #154; stacks on #156.

<details>
<summary>Scope &amp; rationale</summary>

**Why:** the only full-stack CI never builds or loads mobile, so neither #258 (mobile shipped as `resources/`-only) nor the es2015 i18next crash was caught at PR time. The local "~50s saved" is marginal (cached) and not worth a standing footgun plus `if (!SKIP_MOBILE)` branches across the pipeline and gates.

**web-apps (this issue):**
- `.github/workflows/e2e.yml` — drop `SKIP_MOBILE: '1'` → CI builds mobile, and `verify-deploy` then gates it.
- `build/scripts/build-pipeline.js` — remove the conditional; Phase 2 (`deploy-mobile`) always runs. Log a warning if a stray `SKIP_MOBILE` env is set-but-ignored.
- `build/scripts/verify-deploy.mjs` — drop the `if (!SKIP_MOBILE)` wrapper (always assert mobile).
- `build/README.md` — remove the `SKIP_MOBILE` env row.

**DocumentServer follow-ups (not here):** `develop/setup/Makefile` drop the `SKIP_MOBILE` param; optional mobile Playwright test in `e2e/`.

**Depends on #156** — it makes the `deploy-mobile` + `verify-deploy` mobile blocks (which only exist on #156) unconditional, so it stacks on that branch and lands after it.

**Cost:** every CI/local build now builds mobile (~50s first run, cached after). Accepted — the whole point is "always test what ships."
</details>

moodyjmz:
Closes #157.

Removes the `SKIP_MOBILE` escape hatch so CI can no longer skip building and gating mobile — the blind spot that let Euro-Office/DocumentServer#258 ship a broken mobile editor with green CI. Mobile now always builds and `verify-deploy.mjs` always asserts its artifacts.

## Changes (web-apps)

- **`.github/workflows/e2e.yml`** — drop `SKIP_MOBILE: '1'` from the build step. This is the change that matters: it arms the mobile gate #156 added.
- **`build/scripts/build-pipeline.js`** — mobile install + the 4 editor builds + Phase 2 `deploy-mobile` are now unconditional.
- **`build/scripts/verify-deploy.mjs`** — drop the `if (!SKIP_MOBILE)` wrapper; mobile artifacts always asserted.
- **`Readme.md` / `build/README.md` / `docs/webpack-migration.md`** — remove all `SKIP_MOBILE` documentation.

No deprecation shim: the flag was one day old with only two internal call sites (this workflow + the DocumentServer Makefile), so there's no installed base to warn.

## Relationship to #154

#154 Tier 1 (mobile manifest gate) already landed in #156, just gated behind `SKIP_MOBILE`. This PR makes it unconditional and completes #154's **entire web-apps footprint**. #154 Tier 2 (real-browser mobile smoke) lives in `documentserver/e2e` — a separate DocumentServer PR.

## Follow-up (not here — DocumentServer repo)

- `develop/setup/Makefile` still forwards `SKIP_MOBILE`. It's now a silent no-op (harmless — mobile builds regardless), but dead cruft to remove.
- #154 Tier 2 mobile Playwright smoke test.

## Verification

- `git grep SKIP_MOBILE` across web-apps → nothing.
- `node --check` passes on both edited scripts.
- ⚠️ **This is the first time CI will ever build mobile** (every prior e2e run had `SKIP_MOBILE=1`). Watch the build step — if framework7-react has a CI-specific install/build quirk, it surfaces here for the first time.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -55,6 +55,10 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: 20
+          cache: npm
+          cache-dependency-path: |
+            web-apps/build/package-lock.json
+            web-apps/vendor/framework7-react/npm-shrinkwrap.json
 
       # Mirrors web-apps.bake.Dockerfile so the overlay matches the shipped build.
       - name: Install build prerequisites
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -88,7 +88,6 @@ jobs:
           BUILD_ROOT: ${{ github.workspace }}/web-apps/deploy
           THEME: euro-office
           PRODUCT_VERSION: ${{ steps.ver.outputs.product_version }}
-          SKIP_MOBILE: '1'
         run: |
           npm install
           ( cd ../translation && python3 merge_and_check.py )
--- a/Readme.md
+++ b/Readme.md
@@ -101,12 +101,6 @@ docker compose exec eo bash
 export BUILD_NUMBER=0 PRODUCT_VERSION=9.3.2 THEME=euro-office && cd /var/www/onlyoffice/web-apps-develop/build && node scripts/build-pipeline.js
 ```
 
-### Build Flags
-
-| Flag | Description |
-|------|-------------|
-| `SKIP_MOBILE=1` | Skip framework7-react mobile builds (~50s saved) |
-
 ### Environment Variables
 
 | Variable | Description |
--- a/build/README.md
+++ b/build/README.md
@@ -17,10 +17,6 @@ the webpack 5 build system.
 cd /develop/web-apps/build
 PRODUCT_VERSION=9.2.1 BUILD_ROOT=/var/www/onlyoffice/documentserver THEME=euro-office \
   node scripts/build-pipeline.js
-
-# Skip mobile builds (saves ~50s — safe when you haven't touched framework7-react):
-SKIP_MOBILE=1 PRODUCT_VERSION=9.2.1 BUILD_ROOT=... THEME=euro-office \
-  node scripts/build-pipeline.js
 ```
 
 `make web-apps-dev` in the `eo` container calls the same pipeline via the
@@ -45,7 +41,7 @@ Phase 1 — all in parallel
   webpack ×6                  bundle app.js / code.js / app.css / locale per editor (desktop, Terser)
   mobile ×4                   framework7-react webpack builds into the SOURCE tree (word, cell, slide, visio; esbuild)
 
-Phase 2 — mobile deploy (skipped when SKIP_MOBILE=1)
+Phase 2 — mobile deploy
   deploy-mobile.js            copy mobile output (index.html → index.html + index_loader.html, dist/,
                               css/, locale/, resources/img/) from source tree → BUILD_ROOT for the 4
                               mobile editors. MUST precede deploy-theme-images' mobile img overlay.
@@ -59,7 +55,7 @@ Phase 4 — inline
 
 Phase 5 — gates (fail the build loudly)
   verify-bundles.mjs          scan built bundles for surviving {{TOKEN}} literals
-  verify-deploy.mjs           assert every required artifact exists (incl. mobile when !SKIP_MOBILE)
+  verify-deploy.mjs           assert every required artifact exists (incl. mobile)
   verify-browser-target.mjs   no hardcoded browser targets; configs import build/browser-floor.mjs
 ```
 
@@ -73,7 +69,6 @@ Phase 5 — gates (fail the build loudly)
 | `BUILD_ROOT` | **yes** | `../deploy` | Absolute path to DocumentServer output root. |
 | `BUILD_NUMBER` | no | `GITHUB_RUN_NUMBER` → `common.json.build` | Appended to version string in JS bundles. Auto-increments in CI via `GITHUB_RUN_NUMBER`. |
 | `THEME` | no | `default` | Theme directory name under `theme/`. EuroOffice uses `euro-office`. |
-| `SKIP_MOBILE` | no | `0` | Set to `1` to skip framework7-react builds + the Phase 2 mobile deploy (~50s). Safe when not touching mobile code. |
 | `NODE_ENV` | no | `production` | Build mode, forced onto every child. `development` = unminified, `console` kept, esbuild minifier off for mobile (so the ES target + `drop_console` do NOT run — never validate a fix on a dev build). The pipeline echoes the resolved mode in its banner (green prod / red DEV warning). |
 | `WATCH` | no | `0` | `1` = mobile webpack watch mode (live rebuild). Decoupled from `NODE_ENV`; for direct `build.js` runs only — the pipeline would hang on a watching child. |
 | `APP_COPYRIGHT` | no | auto | Copyright line in JS preamble. |
@@ -156,7 +151,7 @@ Preflight + Phase 5. Fail the build loudly rather than ship a silent defect:
 - `verify-replacements.mjs` — preflight: source-side string-replace idiom audit.
 - `verify-bundles.mjs` — scans built bundles for surviving `{{TOKEN}}` literals.
 - `verify-deploy.mjs` — asserts every required artifact exists (vendor/embed/main/forms;
-  mobile `index.html`/`dist`/`css`/`locale` when `!SKIP_MOBILE`).
+  mobile `index.html`/`dist`/`css`/`locale`).
 - `verify-browser-target.mjs` — fails if any webpack/babel/postcss config hardcodes a
   browser target instead of importing from `build/browser-floor.mjs`.
 
--- a/build/scripts/build-pipeline.js
+++ b/build/scripts/build-pipeline.js
@@ -83,6 +83,9 @@ const CHILD_ENV = {
     BUILD_NUMBER,
     THEME,
     NODE_ENV,
+    // Never inherit watch mode from the parent shell — a watching mobile child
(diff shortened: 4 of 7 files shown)


MonaAghili:
# PR #158 Review — Remove SKIP_MOBILE: always build & gate mobile in CI
---

## Summary

The mechanical changes are correct: every `if (!SKIP_MOBILE)` guard is gone, the YAML flag is dropped, docs are cleaned up. The intent is sound. However, this is the **first time CI has ever built mobile**, and that exposes several gaps — two of them capable of silently hanging the pipeline or producing an incomplete deploy with a green exit code.

Seven findings, ranked by severity.

---

## Finding 1 — CONFIRMED / HIGH
### `WATCH=1` from the parent shell leaks into every mobile webpack build → infinite hang

**File:** [`build/scripts/build-pipeline.js:79`](build/scripts/build-pipeline.js#L79)

`CHILD_ENV` is built as:
```js
const CHILD_ENV = { ...process.env, BUILD_ROOT, PRODUCT_VERSION, BUILD_NUMBER, THEME, NODE_ENV };
```
`WATCH` is never explicitly set to `'0'` or deleted. `vendor/framework7-react/build/webpack.config.js` line 56 activates webpack watch mode when `process.env.WATCH === '1'`. If a developer ran the pipeline from a shell where `WATCH=1` was exported (common during a mobile dev session), or if any future CI step exports it, all 4 `mobile:*` spawns enter watch mode and never exit. `phase()` uses `Promise.all` with no timeout — it hangs forever. `e2e.yml` has no `WATCH: '0'` defensive override.

**Failure scenario:** Developer runs `WATCH=1 node scripts/build-pipeline.js` (or a CI runner inherits it) → all 4 `mobile:word/cell/slide/visio` processes enter webpack watch mode → `phase('Phase 1 — parallel')` never resolves → pipeline stalls until GitHub Actions kills the job at 6 hours.

**Fix:**
```js
const CHILD_ENV = {
    ...process.env,
    BUILD_ROOT, PRODUCT_VERSION, BUILD_NUMBER, THEME, NODE_ENV,
    WATCH: '0',   // never inherit watch mode from the parent shell
};
```

---

## Finding 2 — CONFIRMED / HIGH
### No `cache: npm` in `e2e.yml` → 1,123 framework7-react packages downloaded cold on every CI run

**File:** [`.github/workflows/e2e.yml:54`](.github/workflows/e2e.yml#L54)

`setup-node@v4` is configured with only `node-version: 20` — no `cache:` key. `~/.npm` is never restored between runs. The `mobile:install` step inside `build-pipeline.js` runs `npm install --include=dev --production=false` in `vendor/framework7-react/` which has an `npm-shrinkwrap.json` pinning 1,123 packages (1,067 of them devDependencies, ~980 KB lockfile). Every PR run re-downloads all of them cold from the npm registry. This is also the step that blocks all 11 Phase 1 tasks from starting (see Finding 3).

**Failure scenario:** npm registry slowness or a transient network hiccup on the ubuntu-latest runner → cold install times out or yields partial `node_modules` → all 4 mobile webpack builds fail with `Cannot find module 'webpack'` → Phase 1 aborts → build fails before a single desktop bundle is produced.

**Fix:**
```yaml
- uses: actions/setup-node@v4
  with:
    node-version: 20
    cache: npm
    cache-dependency-path: |
      web-apps/build/package-lock.json
      web-apps/vendor/framework7-react/npm-shrinkwrap.json
```

---

## Finding 3 — CONFIRMED / HIGH
### `mobile:install` is `await`ed before `phase('Phase 1 — parallel')` — all 11 non-mobile tasks sit idle while npm installs

**File:** [`build/scripts/build-pipeline.js:264`](build/scripts/build-pipeline.js#L264)

`phase1Tasks` is fully populated (5 deploy scripts + 6 webpack builds) before `mobile:install` is awaited. Phase 1 cannot start until the await resolves:

```js
// phase1Tasks already has 11 tasks here
const install = await runTask(task('mobile:install', 'npm', ...)).promise;  // line 264 — BLOCKS
// ...4 mobile tasks pushed...
const p1 = await phase('Phase 1 — parallel', phase1Tasks);  // line 278 — starts only after install
```

Before this PR, `SKIP_MOBILE=1` in CI meant Phase 1 started immediately. Now every CI run waits for a full npm install before sprites, deploy-common, or any webpack build can emit a byte. On a cold runner with no cache, this is 60–120 s of serialised dead time on the critical path.

**Failure scenario:** Cold CI runner + no cache (see Finding 2) → 90 s npm install → 11 Phase 1 tasks that have zero dependency on mobile sit idle for the entire install duration → wall-clock critical path grows by the full install time with no amortisation across parallel tasks.

---

## Finding 4 — CONFIRMED / MEDIUM
### `deploy-mobile.js` silently skips killed editors, letting Phases 3-4 run against an incomplete `BUILD_ROOT`

**File:** [`build/scripts/deploy-mobile.js:100`](build/scripts/deploy-mobile.js#L100)

When one of the 4 parallel mobile webpack builds fails, `phase()` SIGTERMs the other three. The killed builds never flush `HtmlWebpackPlugin`'s `index.html`. `deploy-mobile` soft-skips any editor whose source `index.html` is absent and exits 0 with no error or warning. Phase 3 (`deploy-resources`, `deploy-theme-images`) and Phase 4 (`inline-svgs`) then run against a partially-populated `BUILD_ROOT`. Only in Phase 5 does `verify-deploy` fail — but by then two more phases of wasted work have run.

**Failure scenario:** `mobile:slide` build fails mid-run → `mobile:word/cell/visio` killed by SIGTERM before `HtmlWebpackPlugin` flushes → `deploy-mobile` skips those 3 editors silently, exits 0 → Phase 3 runs → Phase 4 runs → `verify-deploy` finally exits 1 naming 3 × 6 missing artifacts. The failure is caught, but later than it should be and with a misleading trail.

---

## Finding 5 — CONFIRMED / MEDIUM
### `mobile:install` result is discarded — invisible in the build summary

**File:** [`build/scripts/build-pipeline.js:315`](build/scripts/build-pipeline.js#L315)

```js
const all = [...p0, ...p1, ...pMobile, ...p2, ...p3, ...p4];  // install never spread here
```

The `{ label, ms, code }` object from `runTask(mobile:install)` is used only for the exit-code check and then dropped. The final per-task timing table never shows `mobile:install`. If it runs for 90 s and is the build bottleneck, the summary gives no accounting for that time — the operator sees a wall-clock total inflated by 90 s with no matching task row.

---

## Finding 6 — PLAUSIBLE / MEDIUM
### `framework7.css` and `framework7-rtl.css` are invisible to `verify-deploy`'s CSS check

**File:** [`build/scripts/verify-deploy.mjs:120`](build/scripts/verify-deploy.mjs#L120)

The mobile CSS verification:
```js
for (const [, ref] of html.matchAll(/href="(css\/[^"]+\.css)"/g)) {
    checkFile(`${base}/${ref}`);
}
```
`framework7-bundle.css` and `framework7-bundle-rtl.css` are copied to `css/framework7.css` and `css/framework7-rtl.css` by `CopyWebpackPlugin`. They are loaded via JavaScript (`load_stylesheet('./css/framework7.css')`), not via a static `<link href>` tag. They never appear in `index.html` as `href=` attributes. `checkDir('css/')` passes as long as `app.<hash>.css` exists.

**Failure scenario:** A framework7 version bump renames `framework7-bundle.css` → `CopyWebpackPlugin` source path breaks → `framework7.css` absent from deploy → `verify-deploy` passes (contenthash CSS file still present, regex finds no framework7 href) → mobile editor crashes at runtime for every user loading the stylesheet.

---

## Finding 7 — PLAUSIBLE / LOW
### Issue #157 required a stray-`SKIP_MOBILE` warning; none was implemented

**File:** [`build/scripts/build-pipeline.js:70`](build/scripts/build-pipeline.js#L70) (location of removed constant)

Issue spec item 2 explicitly states: *"Log a warning if a stray SKIP_MOBILE env is set-but-ignored."* `grep SKIP_MOBILE build-pipeline.js` returns nothing — `process.env.SKIP_MOBILE` is never read. The DocumentServer Makefile still forwards `SKIP_MOBILE` (acknowledged in the PR description as dead cruft). A DevOps engineer passing `SKIP_MOBILE=1` to skip mobile in a custom build receives no feedback; mobile runs regardless.

The PR description justifies the omission ("one day old, two internal call sites"), but the requirement is unambiguously unimplemented.

**Fix:**
```js
if (process.env.SKIP_MOBILE) {
    process.stderr.write('warning: SKIP_MOBILE is set but no longer honoured — mobile always builds\n');
}
```

---

## Findings not confirmed

| Candidate | Verdict | Reason |
|-----------|---------|--------|
| `rm('./www/')` race between 4 parallel mobile builds | **Refuted** | `vendor/framework7-react/src/static/` does not exist on disk or in git; `CopyWebpackPlugin` patterns use `noErrorOnMissing: true`; nothing is written to `www/`, making the race inert |
| `NODE_ENV=production` in `CHILD_ENV` causes npm to skip devDeps | **Refuted for CI** | npm 10 (bundled with Node.js 20) honours `--production=false` over `NODE_ENV`. Risk exists only on npm ≤6 (local dev on old runtimes) |
| `Math.max(...[])` → `RangeError` on empty `all[]` | **Refuted (unreachable)** | Every phase either exits the process on failure or returns a non-empty array; `all` can never be empty in the current code paths |

MonaAghili (approved):

PR #162, opened 2026-07-13
Misspell wavy line

pplupo:
Adding the option of the original straight line or a wavy line to underline mispelled words: https://github.com/Euro-Office/DesktopEditors/issues/3

### DIFF
--- a/apps/documenteditor/main/app/controller/LeftMenu.js
+++ b/apps/documenteditor/main/app/controller/LeftMenu.js
@@ -532,6 +532,9 @@ define([
                     value = Common.localStorage.getBool("de-spellcheck-ignore-numbers-words", true);
                     Common.Utils.InternalSettings.set("de-spellcheck-ignore-numbers-words", value);
                     spprops.put_IgnoreWordsWithNumbers(value);
+                    value = Common.localStorage.getBool("de-spellcheck-wavy-line", true);
+                    Common.Utils.InternalSettings.set("de-spellcheck-wavy-line", value);
+                    spprops.put_WavyLine(value);
                     this.api.asc_setSpellCheckSettings(spprops);
                 }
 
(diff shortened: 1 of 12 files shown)


pplupo:
This is also required: https://github.com/Euro-Office/sdkjs/pull/59

Alex-Arsys (commented):
1 · locale — missing i18n key — blocking (convention)
The new label adds strSpellcheckWavyLine as a JS default, but I don't see the corresponding DE.Views.FileMenuPanels.Settings.strSpellcheckWavyLine key in apps/documenteditor/main/locale/en.json (the sibling strIgnoreWordsWithNumbers is there). As-is the checkbox label shows in English via the JS fallback but stays untranslated in every locale. Could you add the key to en.json (and flag the other locales for translation) to match the project convention?

2 · scope — DE only vs SSE/PE — question / blocking
The rendering change in sdkjs#59 lives in the common engine (Graphics.js, IsSpellCheckWavyLine defaults to true), so Spreadsheet and Presentation will render the wavy underline by default — but this PR only wires the toggle into DocumentEditor. Is DE-only intended (the source issue is DesktopEditors), or should we replicate the checkbox in SSE/PE so users there can turn it off too?

3 · cross-repo dependency — non-blocking
Note this depends on put_WavyLine from sdkjs#59; if web-apps ships against an sdkjs without it, spprops.put_WavyLine is undefined and applying spell-check settings throws. Worth merging / deploying the two together and cross-linking the PRs.

pplupo:
@Alex-Arsys thank you for pointing those out. You are absolutely right about all of them.

1- Fixed.

2- When I implemented it, I initially only thought about DE. However, you are absolutely right that it makes sense to have these in all other editors, so my fix wasn't to make it DE-only, but to implement it in the others as well. With one caveat: SSE doesn't have any sort of underline (wavy or straight). Or, at least, I didn´t know how to trigger it. I tried in a cell, in a cell comment, in a textbox, and inside a shape. There's no misspelling underline of any kind. So I just removed the settings from SSE.

moodyjmz:
This e2e failure isn't a bug in this PR — it's failing because the sdkjs API it calls (`put_WavyLine`) doesn't exist yet in the engine. The companion change ([sdkjs#59](https://github.com/Euro-Office/sdkjs/pull/59), same author/branch) is still open and unmerged, so the nightly image this PR gets tested against doesn't have it. Once #59 merges and the nightly image picks it up, this should go green with no changes needed here.

PR #167, opened 2026-07-18
fix: collapse empty icon svgs inflating language menus (#165)

### LINKED ISSUE #165: Language selection dropdown mangled by large empty svgs
Most language entries are vertically expanded by empty svgs:

Image

After deleting the svg from the DOM:

Image

Maybe related to changes from https://github.com/Euro-Office/web-apps/issues/106? These svgs are presumably meant to be images in line with the images visible to the left of "English -- UK" and "English -- Canada", but instead break formatting. The task is to figure out what these should look like, and replace these with properly rendering images (or svgs, etc).

cc @moodyjmz

moodyjmz:
Closes #165.

## Problem

Rows in the "Set document language" dropdown were vertically inflated by tall empty `<svg>` elements. The reporter guessed the svgs "should be flag images" — they should not: the icon is the spell-check indicator, and non-spell-check languages are meant to show *no* icon.

## Root cause

Commit 4fe48ff7 replaced the old empty `<i class="icon">` placeholder with `<svg class="icon uni-scale">` carrying a *conditional* `<use>` child. When `spellcheck` is false the template emits a literal `<svg class="icon uni-scale"></svg>`.

An empty inline `<svg>` with no width/height attribute and no CSS sizing in its context falls back to the CSS replaced-element intrinsic size — **300×150px**. The old `<i>` collapsed to zero when empty; the svg does not. `svg.icon` gets sizing only from context selectors (button/dataview contexts, `.spellcheck-lang`, etc.); anything outside one hits the fallback.

## Fix

- **`bigscaling.less`** — a single global rule `svg.icon:empty, svg.menu-item-icon:empty { display:none }`. This neutralises the whole empty-svg class: it fixes the language combo dialog **and** the identical twin in the document-editor status-bar language menu (`Statusbar.js`) with no per-template edits. Standardised, not sprinkled.
- **`language-dialog.less`** — a scoped rule hides the combo *input* svg when `.spellcheck-lang` is toggled off (that svg keeps its `<use>`, so `:empty` cannot reach it — a distinct "unsized non-empty svg" case).

## Verification

- Headless Chrome, measured against the *unmodified* templates: non-spell-check rows 191px → 38px; the status-bar `.lang-menu` collapses identically.
- Live in the running editor: the status-bar language menu inflated rows (180px) collapse to 42px, spell-check icons intact.

## Not covered (follow-up)

Two `.toolbar__icon`-container mode-B suspects (`PluginDlg.js`, `Tab.js`) have a live `<use>` and so are not `:empty`; they need context-specific sizing and in-browser confirmation. Out of scope here.

### DIFF
--- a/apps/common/main/resources/less/bigscaling.less
+++ b/apps/common/main/resources/less/bigscaling.less
@@ -76,9 +76,24 @@
 // row. An empty icon svg is never legitimate (nothing to paint), so collapse it
 // globally — this covers the non-spell-check rows in the language combo and
 // status-bar language menus, and any future empty-svg template.
+//
+// !important is deliberate: this is an invariant ("nothing to paint" is never
+// legitimately visible), not a default meant to be overridable. Without it,
+// higher-specificity or later-imported rules elsewhere (e.g. toolbar.less's
+// `svg.icon { display: block }` context rules) silently win and reintroduce
+// the bug — see https://github.com/Euro-Office/web-apps/issues/169.
 svg.icon:empty,
 svg.menu-item-icon:empty {
-    display: none;
+    display: none !important;
+}
+
+// :empty only matches when a node has zero children, including whitespace —
+// so this rule alone is fragile to template reformatting (see
+// https://github.com/Euro-Office/web-apps/issues/168). Known conditional-icon
+// templates (Statusbar.js, LanguageDialog.js) additionally set this class
+// explicitly, so their empty state doesn't depend on whitespace-free markup.
+.icon-hidden {
+    display: none !important;
 }
 
 .btn-category,
--- a/apps/common/main/resources/less/bigscaling.less
+++ b/apps/common/main/resources/less/bigscaling.less
@@ -70,6 +70,17 @@
     }
 }
 
+// An <svg class="icon"> whose <use> lives inside a template conditional renders
+// empty when the condition is false. An empty inline svg has no width/height, so
+// it falls back to the 300x150 intrinsic replaced-element size and inflates its
+// row. An empty icon svg is never legitimate (nothing to paint), so collapse it
+// globally — this covers the non-spell-check rows in the language combo and
+// status-bar language menus, and any future empty-svg template.
+svg.icon:empty,
+svg.menu-item-icon:empty {
+    display: none;
+}
+
 .btn-category,
 .btn-options,
 .combo-dataview,
--- a/apps/common/main/resources/less/language-dialog.less
+++ b/apps/common/main/resources/less/language-dialog.less
@@ -47,7 +47,7 @@ li {
   // toggles .spellcheck-lang to show/hide it. Without that class the SVG has
   // no sizing and falls back to the 300x150 intrinsic size, so collapse it
   // (the old <span>+sprite icon collapsed to nothing when unsized).
-  svg.input-icon:not(.spellcheck-lang):not(.lang-flag) {
+  svg.input-icon:not(.spellcheck-lang) {
     display: none;
   }
 
--- a/apps/common/main/resources/less/language-dialog.less
+++ b/apps/common/main/resources/less/language-dialog.less
@@ -43,6 +43,14 @@ li {
 }
 
 .combo-langs {
+  // The spell-check indicator SVG always carries its <use>; onLangSelect
+  // toggles .spellcheck-lang to show/hide it. Without that class the SVG has
+  // no sizing and falls back to the 300x150 intrinsic size, so collapse it
+  // (the old <span>+sprite icon collapsed to nothing when unsized).
+  svg.input-icon:not(.spellcheck-lang):not(.lang-flag) {
+    display: none;
+  }
+
   .dropdown-menu {
     li .icon.lang-flag {
       margin-top: 1px;
--- a/apps/common/main/lib/view/LanguageDialog.js
+++ b/apps/common/main/lib/view/LanguageDialog.js
@@ -95,7 +95,7 @@ define([], function () { 'use strict';
                 '<li id="<%= id %>" data-value="<%= value %>">',
                     '<a tabindex="-1" type="menuitem" langval="<%= value %>">',
                         '<div>',
-                            '<svg class="icon uni-scale<% if (spellcheck) { %> spellcheck-lang<% } %>"><% if (spellcheck) { %><use href="#btn-ic-docspell"></use><% } %></svg>',
+                            '<svg class="icon uni-scale<% if (spellcheck) { %> spellcheck-lang<% } else { %> icon-hidden<% } %>"><% if (spellcheck) { %><use href="#btn-ic-docspell"></use><% } %></svg>',
                             '<%= displayValue %>',
                         '</div>',
                         '<label style="opacity: 0.6"><%= displayValueEn %></label>',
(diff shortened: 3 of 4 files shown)


apps/common/main/resources/less/bigscaling.less
```diff
@@ -70,6 +70,17 @@
...
 }
 
+// An <svg class="icon"> whose <use> lives inside a template conditional renders
+// empty when the condition is false. An empty inline svg has no width/height, so
+// it falls back to the 300x150 intrinsic replaced-element size and inflates its
+// row. An empty icon svg is never legitimate (nothing to paint), so collapse it
+// globally — this covers the non-spell-check rows in the language combo and
+// status-bar language menus, and any future empty-svg template.
+svg.icon:empty,
+svg.menu-item-icon:empty {
```
MonaAghili, line 86: Confirmed correct and sufficiently scoped, verified against every other `svg.icon`/`svg.menu-item-icon` producer in the codebase (`MenuItem.js`, `Button.js`, `DataView.js`, spreadsheet/presentation status bars), none of which are conditionally empty. Two notes for a follow-up, not this PR:

1. `:empty` only matches zero child *nodes*, including whitespace. A future multi-line reformat of the template in `Statusbar.js:262` or `LanguageDialog.js:98` would silently break this match with no error.
2. This file reads as scaling-specific, so a general icon-visibility rule living here may not be where someone looks first when debugging a similar issue later.

apps/common/main/resources/less/language-dialog.less
```diff
@@ -43,6 +43,14 @@ li {
 }
 
 .combo-langs {
+  // The spell-check indicator SVG always carries its <use>; onLangSelect
+  // toggles .spellcheck-lang to show/hide it. Without that class the SVG has
+  // no sizing and falls back to the 300x150 intrinsic size, so collapse it
+  // (the old <span>+sprite icon collapsed to nothing when unsized).
+  svg.input-icon:not(.spellcheck-lang) {
```
MonaAghili, line 50: Correct and necessary. This svg (`LanguageDialog.js:86`) always contains a `<use>`, so `:empty` alone would never catch it; `onLangSelect` only toggles the `spellcheck-lang` class. Good call keeping this as its own scoped rule instead of trying to fold it into the global one.

MonaAghili:
Did a second, more skeptical pass on top of the initial review, checking presentation/spreadsheet editors for the same bug pattern, RTL, dark mode/print, and specificity conflicts against every other `display:` rule on `svg.icon`/`svg.menu-item-icon` in the codebase.

**Confirmed correct and complete for the reported bug:**

- The doc-editor status bar (`Statusbar.js:262`) and language dialog (`LanguageDialog.js:98`) are the *only two* conditionally-empty-svg sites in the app. Verified the presentation editor's status bar uses `<i class="icon spellcheck">` (not svg, so it never had this bug) and the spreadsheet editor's `LeftMenu.template` always renders `<use>` unconditionally. No missed cases.
- RTL positioning rules only touch the always-populated `input-icon`/`spellcheck-lang` element, never the collapsed rows. No conflict there.
- No dark-mode/print/sibling-selector rules reference these classes anywhere, so no risk on that front.

**One new finding worth a follow-up ticket (not a blocker):** `apps/documenteditor/main/resources/less/toolbar.less:98-101` has `.menu-pageposition .dataview svg.icon { display: block; }` at specificity `(0,3,1)`, higher than the new `svg.icon:empty` rule at `(0,2,1)`. Today this is harmless since that dataview's icons are always populated (never conditionally empty), but if that template ever becomes conditional (the exact pattern that caused this bug), the new global rule won't save it there. This higher-specificity rule wins regardless of import order. Also noticed `toolbar.less`'s `.item-shape`/`.item-chartlist`/etc. `svg.icon { display: inline-block }` rules are the *same* specificity as the new rule and only lose today because of import order in `app.less`, not specificity. Reordering imports later would silently reintroduce the collision.

Approving on the basis that this fixes the reported bug correctly with no active regressions found. Filing separately:
1. Fix the template layer so `<svg class="icon">` is never emitted without a `<use>` in the first place, since the CSS rule is currently the only defense.
2. Add a regression test (row height / svg empty-state) for the language dialog and status bar.
3. Note the specificity fragility above so nobody "cleans up" the import order in `app.less` without knowing it's load-bearing.

---

## `apps/documenteditor/main/resources/less/toolbar.less:100`
*(the `display: block;` line inside `.menu-pageposition .dataview svg.icon`)*

Flagging for awareness, not blocking this PR: this rule's specificity `(0,3,1)` beats the new `svg.icon:empty` rule in `bigscaling.less` `(0,2,1)`. Currently harmless since this dataview's icons are always populated, but if this template ever becomes conditionally-empty (the pattern that caused the bug this PR fixes), this rule would silently win and the row-inflation bug would resurface here, uncaught by the new global fix.

apps/common/main/resources/less/language-dialog.less
moodyjmz: Thanks, that's the human input there

apps/common/main/resources/less/bigscaling.less
moodyjmz: 2. I wondered that too tbh

moodyjmz:
@MonaAghili added a follow-up commit hardening the two fragility points you flagged in review — `!important` on the `:empty` collapse rule (specificity/import-order fragility) and an explicit `icon-hidden` class in `Statusbar.js`/`LanguageDialog.js` (whitespace fragility of `:empty` itself). Refs #168, #169.

Could you take another look and convert to an approval if it holds up?

MonaAghili:
## Review: collapse empty icon svgs inflating language menus

**Verdict: solid fix, already hardened from an earlier review round.** This branch has 3 commits — the original fix, a follow-up for the combo input svg, and a hardening commit (`6ac1c64`) explicitly responding to specificity/fragility issues raised in a prior review (which spawned follow-up issues #168 and #169, both confirmed open).

**What it does:** an empty `<svg class="icon">` (conditional `<use>` not rendered) has no intrinsic size, so it falls back to the CSS replaced-element default 300×150px, inflating dropdown rows. Fix collapses empty icon svgs globally via `svg.icon:empty, svg.menu-item-icon:empty { display: none !important }` in `bigscaling.less`, plus an explicit `.icon-hidden` class set by the two known conditional templates (`Statusbar.js`, `LanguageDialog.js`) as a non-`:empty`-dependent backstop, plus a scoped rule for the combo's spell-check input icon (which always has a `<use>`, so `:empty` can't reach it).

**Verified independently:**
- `apps/documenteditor/main/resources/less/toolbar.less:98-101` really does have `.menu-pageposition .dataview svg.icon { display: block }` — genuinely higher specificity (0,3,1) than the new rule (0,2,1), and imported *after* `bigscaling.less` in `app.less` (line 140 vs 130), so the `!important` is load-bearing, not decorative.
- Several `svg.icon { display: inline-block }` rules in the shared `toolbar.less` (`.item-shape`, `.item-chartlist`, etc.) are equal specificity to the new rule and only lose today due to import order — confirms the second `!important` justification too.
- The combo input svg (`LanguageDialog.js:86`) always renders with a `<use>` and toggles `.spellcheck-lang` via JS (`:147`) — matches the `svg.input-icon:not(.spellcheck-lang)` rule added in `language-dialog.less`.
- No pre-existing `.icon-hidden` class collision anywhere else in the codebase.

**Minor, non-blocking:**
- No automated test coverage — consistent with this codebase's existing lack of CSS/layout unit tests; verification is manual/headless-Chrome measurement per the PR description.
- `.icon-hidden` and the `:empty` selector are redundant for the two known templates by design (defense-in-depth), not a bug.
- Follow-up work (auditing other conditional-icon templates, `PluginDlg.js`/`Tab.js`) is correctly scoped out and already tracked separately rather than scope-creeping this PR.

Issue #171, opened 2026-07-21
Dependabot triage: dismiss 37 stale alerts, track 6 live build-toolchain fixes


moodyjmz:
## Summary

The Dependabot security overview reports **43 open alerts** (3 critical / 24 high / 13 medium / 3 low). Triage shows **37 are stale** — they point at dependencies no longer in the tree, or at a manifest that was deleted from `main` — and **6 are genuinely live**, all build/test-time toolchain transitives with no shipped-runtime or untrusted-input path.

This issue records the decision to **dismiss the 37 stale alerts** and **track the 6 live ones for a fix**, with the evidence behind each call.

All analysis was read-only (`gh api`, `git show origin/main:…`); classification tested the **installed version against each advisory's vulnerable range** with semver, not by package-name presence.

## Scope check

- Every alert is under `build/` — the webpack/mocha build toolchain. The shipped browser code (`apps/…`) is **produced by** this toolchain, not dependent on it; none of the flagged packages are bundled into output.
- `vendor/backbone` and `vendor/underscore` lockfiles carry **zero** alerts in any state.
- Automated security updates are currently **disabled** and there is no `.github/dependabot.yml` — so no fix PRs were ever raised (tracked as follow-up C below).

## Kept open — 6 live (fix in follow-up B)

| Alert | Sev | Package (installed) | Advisory / CVSS | Patched | Pulled in by |
|---|---|---|---|---|---|
| `#37` | high | `@babel/plugin-transform-modules-systemjs` 7.25.0 | GHSA-fv7c-fp4j-7gwp / 8.2 | 7.29.4 | @babel/preset-env |
| `#19` | high | `picomatch` 2.3.1 | GHSA-c2c7-rcm5-vvqj / 7.5 | 2.3.2 | chokidar → readdirp/anymatch |
| `#20` | med | `picomatch` 2.3.1 | GHSA-3v7f-55p6-f55p / 5.3 | 2.3.2 | chokidar → readdirp/anymatch |
| `#48` | med | `js-yaml` 4.1.0 | GHSA-h67p-54hq-rp68 / 5.3 | 4.2.0 | mocha |
| `#38` | med | `serialize-javascript` 6.0.0/6.0.2 | GHSA-qj8w-gfj5-8c6v / 5.9 | 7.0.5 | copy-webpack-plugin, terser |
[...]

PR #173, opened 2026-07-21
fix(build): pin picomatch >=2.3.2 to clear ReDoS advisory (#19/#20)

moodyjmz:
## What
Pin `picomatch` to `>=2.3.2` in `build/` via a version-scoped npm `overrides` entry, clearing two Dependabot alerts.

## Why
`picomatch@2.3.1` is pulled transitively through chokidar's `anymatch (^2.0.4)` / `readdirp (^2.2.1)` (webpack file-watch). It is vulnerable to:
- **GHSA-c2c7-rcm5-vvqj** (CVE-2026-33671, CVSS 7.5, high) — Dependabot `#19`
- **GHSA-3v7f-55p6-f55p** (CVSS 5.3, medium) — Dependabot `#20`

## How
```json
"picomatch@<2.3.2": "2.3.2"
```
Version-scoped so only the vulnerable 2.x line is bumped — a patch, within the `^2` ranges of `anymatch`/`readdirp`. tinyglobby's already-patched `picomatch@4.0.4` is deliberately left untouched. Follows the existing `overrides` pattern in `build/package.json`. Build-time dependency only; not shipped to the browser.

## Validation
- `npm ci` clean (EXIT 0) on the regenerated lockfile.
- Resolved tree: `picomatch 2.3.2` (top-level) + `4.0.4` (tinyglobby, unchanged); no `picomatch <2.3.2` remains.
- Diff is 1 override line + the lockfile's picomatch node.

Part of the Dependabot backlog triage — refs #171. Does **not** address `#38` (serialize-javascript), which is tracked separately (a forced 6→7 major on `copy-webpack-plugin` for a non-runtime-reachable medium).

### DIFF
--- a/build/package-lock.json
+++ b/build/package-lock.json
@@ -4149,10 +4149,11 @@
       "license": "ISC"
     },
     "node_modules/picomatch": {
-      "version": "2.3.1",
-      "resolved": "https://registry.npmjs.org/picomatch/-/picomatch-2.3.1.tgz",
-      "integrity": "sha512-JU3teHTNjmE2VCGFzuY8EXzCDVwEqB2a8fsIvwaStHhAWJEeVd1o1QD80CU6+ZdEXXSLbSsuLwJjkCBWqRQUVA==",
+      "version": "2.3.2",
+      "resolved": "https://registry.npmjs.org/picomatch/-/picomatch-2.3.2.tgz",
+      "integrity": "sha512-V7+vQEJ06Z+c5tSye8S+nHUfI51xoXIXjHQ99cQtKUkQqqO1kO/KCJUfZXuB47h/YBlDhah2H3hdUGXn8ie0oA==",
       "dev": true,
+      "license": "MIT",
       "engines": {
         "node": ">=8.6"
       },
@@ -4886,7 +4887,6 @@
       "resolved": "https://registry.npmjs.org/terser/-/terser-5.48.0.tgz",
       "integrity": "sha512-J/9An6vs9Us6wKRriSFXBWdRZapREHqFzdNUKk0pmu804EMR6dr6winwo7e5JDxN4xahxQsuysyYFwlwj4XN/Q==",
       "license": "BSD-2-Clause",
-      "peer": true,
       "dependencies": {
         "@jridgewell/source-map": "^0.3.3",
         "acorn": "^8.15.0",
--- a/build/package.json
+++ b/build/package.json
@@ -25,7 +25,8 @@
     "semver-regex": "3.1.3",
     "trim-newlines": "3.0.1",
     "mkdirp": "0.5.6",
-    "requirejs": "2.3.7"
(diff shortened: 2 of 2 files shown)


MonaAghili (approved):
**Approved.** The `build/` fix has been independently verified: the version-scoped override resolves correctly and both referenced CVEs are cleared in that manifest.

**Non-blocking follow-ups:**

1. **Override lifecycle.** `"picomatch@<2.3.2": "2.3.2"` is only effective while `mocha`/`chokidar` continue pulling picomatch 2.x transitively. Once that dependency chain moves to 3.x or 4.x, the key will stop matching any package with no error or warning, and will remain as unused configuration. Recommend tracking this with a follow-up issue (or a note in the commit trail) so it gets removed once it's no longer needed.

2. **Vendor lockfiles not covered by this PR.** `vendor/underscore/package-lock.json` and `vendor/backbone/package-lock.json` still resolve `picomatch@2.3.1`, affected by the same two CVEs. Recommend a follow-up PR to bring these in line, so the vulnerability isn't considered fully remediated for the repository while it remains open in these manifests.

<details>
<summary>Verification details</summary>

- **Override correctness / CVE remediation**: checked out this PR's commit and ran `npm ci` twice (clean install, byte-identical lockfile both times), `npm ls picomatch` (resolves to `2.3.2` on the anymatch/readdirp branch; `4.0.4` under tinyglobby is unaffected), and `npm audit` (no picomatch findings). Cross-referenced both advisories directly — GHSA-c2c7-rcm5-vvqj and GHSA-3v7f-55p6-f55p both list `2.3.2` as the patched version.
- **Override lifecycle claim**: consistent with documented npm override semantics — a version-scoped key (`"picomatch@<2.3.2"`) only applies to resolved instances matching that range. Once no instance in the tree resolves below 2.3.2, the override becomes a no-op with no diagnostic.
- **Vendor lockfile exposure**: confirmed directly against this PR's commit; both `vendor/underscore` and `vendor/backbone` lockfiles resolve `picomatch@2.3.1`, unchanged by this PR.

</details>

moodyjmz:
Thanks for the thorough review, @MonaAghili — verified both follow-ups:

- **Vendor lockfiles:** confirmed. `vendor/underscore` and `vendor/backbone` both pin `picomatch@2.3.1` (same two advisories), untouched by this PR. Good catch — it also corrected our earlier triage note that treated those manifests as "clean" (they're *unscanned*, not clean). Tracked in #175, with the reachability caveat that neither lockfile is installed or executed by the build — underscore/backbone ship pre-minified via webpack aliases, and their locks are the libraries' own dev/test harness — so it's on-paper rather than runtime-reachable.
- **Override lifecycle:** agreed. The version-scoped override becomes a silent no-op once the tree moves off picomatch 2.x; noted for removal when chokidar's anymatch/readdirp advance.

Merging on your approval — thanks again.

PR #179, opened 2026-07-23
chore(build): remove dead grunt-inline plugin

moodyjmz:
Removes the vendored `grunt-inline` plugin — unreferenced since Phase E removed grunt; `build/scripts/inline-svgs.js` has already replaced its functionality.


j-base64 (approved):
Look good since grunt's already gone. Just to double check, searched for references to grunt-inline in files, found none except a comment in inline-svgs.js explaining the replacement. So confident to approve. LGTM

MonaAghili (approved):

PR #182, opened 2026-07-27
Add icon to mobile close action

Phillipxh:
## What changed

- add the shared close icon to the mobile close menu item
- apply the change to document, spreadsheet, presentation, and Visio editors
- remove the legacy 36 px title indentation now that the item has a media icon

## Why

The close action under Feedback & Support was the only menu action without an icon. Its placeholder indentation also pushed the label farther right than neighboring items.

## Validation

- production mobile build: document editor
- production mobile build: spreadsheet editor
- production mobile build: presentation editor
- production mobile build: Visio editor
- `git diff --check`

### DIFF
--- a/apps/common/mobile/resources/less/common.less
+++ b/apps/common/mobile/resources/less/common.less
@@ -1249,15 +1249,10 @@ input[type="number"]::-webkit-inner-spin-button {
 
 // Close editor button
 .close-editor-btn {
-    .item-inner {
-        padding-left: 36px;
-    }
-
     .item-title {
         color: @text-error;
     }
 }
-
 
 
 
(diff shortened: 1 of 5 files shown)


j-base64:
Hi @Phillipxh , thanks for your contrib! Nice cleanup on the padding, it matches how the other icon rows in this list already get their spacing.

One thing worth a look: IconClose only has an Android asset (@common-android-icons/icon-close.svg), so right now look like it will renders that same glyph on iOS too. Everywhere else this codebase uses IconClose, it's gated behind Device.android &&, with iOS showing text instead ("Cancel").

Could you have a look onto this and maybe gate your change on the same way as the other IconClose call sites do to follow the android/ios inplace best practice? (icon on Android, text on iOS), or add an iOS-styled asset if justified

Phillipxh:
Thanks for spotting the platform convention. I pushed commit `f83934ce1b`, which gates the close icon with `Device.android` in all four mobile editors. iOS therefore keeps the text-only row, consistent with the other `IconClose` call sites.

All four production mobile bundles (word, cell, slide, and visio) compile successfully.

PR #183, opened 2026-07-27
feat(theme): add per-editor desktop accent strip

Phillipxh:
## Summary

- add a narrow 8px identity-colored strip across the top of each desktop editor
- keep the existing neutral header, icons, controls, ribbon, and canvas unchanged
- give PDF forms the PDF accent even though they run inside the document editor
- preserve the existing behavior of light, dark, night, and high-contrast themes

## Motivation

A restrained accent gives each editor some visual identity and context without applying a large block of color to the full header. This follows the design direction suggested in review and is similar to the narrow application strip used by the Collabora integration.

## Implementation

- define five Euro Office accent color tokens for document, spreadsheet, presentation, PDF, and drawing
- render the accent as an 8px `border-top` on the full `#viewport`, using `box-sizing: border-box` so no editor content is covered
- override the document accent with the PDF accent for `body.pdf-form`
- retain the pre-existing per-editor header logo rules unchanged

## Impact

The only new visual treatment is the narrow application accent strip. Header foreground/background tokens and the global SVG icon palette are no longer overridden, so every theme keeps its existing contrast behavior.

## Validation

- production webpack build: document editor
- production webpack build: PDF editor
- standalone Euro Office LESS compilation
- verified generated CSS contains all editor accents and the PDF-form override
- browser-target consistency check
- `git diff --check`


Phillipxh:
Hi @moodyjmz and @Alex-Arsys, could you please review this PR when you have a chance?

I have rebased it onto the latest `main`. The Euro Office LESS theme compiles successfully, the editor header color tokens are present in the generated CSS, and the browser-target consistency check passes. The PR is currently mergeable and the DCO check is green.

Thank you!

moodyjmz:
Sorry this sat for so long.

Had a proper look now. I pulled some before/after grabs against a local 9.3.4 build so I could judge it in context rather than just reading the diff:

compare-light

compare-dark

I think it looks good. The coloured header reads far better than the neutral one, and the icon context bridge is a tidier solution than I was expecting.

Three things before it goes in. One is a definite change, one I think is worth doing, and the third is an open question I don't think either of us should decide on our own.

### The active tab indicator disappears in the light themes

`--highlight-header-tab-underline-*` is set to the brand colour, and that's now the header background too. In `theme-white` it's an exact match for document, PDF and visio, so "which tab am I on" goes away entirely:

compare-underline-fix


Upstream hit this in the classic theme and pointed those tokens at the header foreground instead (`colors-table.less:84-88`). Same trick works here — five lines in the block you've already got. The third panel above is that patch running locally, not a mock-up:

```diff
   --text-toolbar-header: @eo-header-foreground;
   --icon-toolbar-header: @eo-header-foreground;
 
+  --highlight-header-tab-underline-document: @eo-header-foreground;
+  --highlight-header-tab-underline-spreadsheet: @eo-header-foreground;
+  --highlight-header-tab-underline-presentation: @eo-header-foreground;
+  --highlight-header-tab-underline-pdf: @eo-header-foreground;
+  --highlight-header-tab-underline-visio: @eo-header-foreground;
```

The dark themes aren't affected, incidentally — they draw the active tab as a filled block rather than an underline, so it survives either way.

### Two of the colours drop below AA

White on the header backgrounds:

| | this PR | ratio | upstream value it replaces |
|---|---|---|---|
| document | `#446995` | 5.68:1 | `#446995`, unchanged |
| **spreadsheet** | `#40865c` | **4.39:1** | `#3A8056` — 4.77:1 |
| **presentation** | `#be664f` | **4.04:1** | `#B75B44` — 4.57:1 |
| pdf | `#aa5252` | 5.20:1 | `#AA5252`, unchanged |
| visio | `#444796` | 8.12:1 | `#444796`, unchanged |

The header labels are normal-weight 11–12px, so 4.5:1 is the bar. What's awkward is that spreadsheet and presentation are the only two colours the PR actually changes — the other three are byte-identical to upstream — and they're exactly the two that regress. That's because mobile and desktop disagree on those two specifically, and the PR takes mobile's values for all five.

Going back to `#3A8056` and `#B75B44` fixes AA and shrinks the diff to almost nothing for the classic theme. Not a hill I'd die on if there's a reason for the mobile values, but I'd want to hear it.

### Should the header ignore the theme at all?

`:root body[data-editor-type]` has no theme qualifier, so it overrides dark, night and contrast-dark as well as the light ones. The second grab above is dark mode: blue header, dark ribbon. `theme-contrast-dark` is the accessibility theme, so that one in particular I'm uneasy about.

Worth noting mobile doesn't do this — it retints the whole brand set in dark mode and flips the foreground to black (`colors-table-dark.less:19-26`). So "match mobile" arguably argues for a dark variant rather than one palette everywhere.

No strong view on which way this should go, but it's a design decision rather than a code one and I don't want to be the only person signing it off. I'll ask for some design input on this and update.

Thanks for your patience, and sorry again for the wait.

<details>
<summary>Full-frame grabs</summary>


after-theme-dark-full

after-theme-white-full

before-theme-dark-full

before-theme-white-full

</details>

moodyjmz (changes requested):
@Phillipxh Hiya, I left a bit of feedback in a comment and will update once I hear back from design

https://github.com/Euro-Office/web-apps/pull/183#issuecomment-5688430059

Phillipxh:
Thanks for the detailed review and the comparison screenshots. I have pushed commit `78c35bc378` addressing the two actionable points:

- all five `--highlight-header-tab-underline-*` tokens now use `@eo-header-foreground`, so the active tab remains visible in light themes;
- spreadsheet and presentation backgrounds now use the AA-compliant upstream shades `#3a8056` and `#b75b44`.

I verified the Euro Office LESS output and the browser-target consistency check. I have deliberately left the dark/night/contrast-dark behavior unchanged while awaiting the design input you mentioned.

moodyjmz:
Thanks for turning that around so quickly — underline and the two colour values both look right to me.

compare-updated

Checked at 78c35bc against a local 9.3.4 build. The tab indicator is back in all three coloured headers, and spreadsheet and presentation now measure 4.77:1 and 4.57:1, so both clear AA.

### PDF comes out two-tone

This is the one still outstanding, and it's a bit more tangled than I made it sound last time — I got it wrong first go, so here are the measured values rather than my impression.

Opening a PDF gives you `data-editor-type="document"` on the body plus a `pdf-form` class; it runs in the document editor, not the PDF one. The awkward part is that different bits of the chrome then read different tokens:

| element | reads | now paints |
|---|---|---|
| `.toolbar` | `--toolbar-header-document` | `rgb(68,105,149)` — the document blue |
| `.toolbar .extra` | `--toolbar-header-pdf` | `rgb(243,243,243)` — neutral |
| `#box-document-title` | `--toolbar-header-pdf` | `rgb(243,243,243)` — neutral |

`.extra` and the title box go through the `.pdf-form &` branch of `.header-background-color()` (`asc-mixins.less:839-845`), which points at `--toolbar-header-pdf`. Since `body[data-editor-type="pdf"]` never matches here, that token stays at the theme default. So you get a blue toolbar with a neutral strip across it — the pale band in the grab above is that, not a bad screenshot.

Setting only `--toolbar-header-pdf` isn't enough either; I tried it and it just flips the mismatch the other way (red strip, blue toolbar). Both have to agree:

```less
:root body.pdf-form {
  --toolbar-header-document: @eo-header-pdf-background;
  --toolbar-header-pdf: @eo-header-pdf-background;
  --text-toolbar-header-on-background-pdf: @eo-header-pdf-avatar-text;
}
```

That gives `rgb(170,82,82)` on all three, which is the last row in the grab.

Caveat: I only had a form PDF to hand (`sample.pdf`), so this took the `pdf-form` path. A plain non-form PDF may well load the pdfeditor app where the attribute does match, and that would want handling separately — worth a look before settling on this.

### Should the header ignore the theme at all?

`:root body[data-editor-type]` has no theme qualifier, so it overrides dark, night and contrast-dark as well as the light ones:

compare-dark-updated

`theme-contrast-dark` is the accessibility theme, so that one in particular I'm uneasy about.

Worth noting mobile doesn't do this — it retints the whole brand set in dark mode and flips the foreground to black (`colors-table-dark.less:19-26`). So "match mobile" arguably argues for a dark variant rather than one palette everywhere.

No strong view on which way this should go, but it's a design decision rather than a code one and still waiting on that.

Thanks again for the quick turnaround.

Phillipxh:
Implemented the PDF form header fix in commit `70075c95a7`.

`body.pdf-form` now sets both `--toolbar-header-document` and `--toolbar-header-pdf` to `@eo-header-pdf-background`, plus the PDF avatar-text token. This keeps the toolbar, extra section, and document title on the same PDF identity color while leaving the existing `body[data-editor-type="pdf"]` rule in place for the standalone PDF editor.

Validation:
- Euro Office LESS compiles successfully
- generated PDF-form tokens resolve to `#aa5252`, `#aa5252`, and `#8d4444`
- browser-target consistency check passes
- `git diff --check` passes

marcoambrosini (commented):
Hi, thanks for your contribution!
from my point of view this is too much color and we should dial this down a bit. With Collabora we adopted a colored strip above the whole window to give each app some flavor and some more visual context about where which app the user has opened. I think we should do the same here.

Image

Phillipxh:
@marcoambrosini @moodyjmz I have reworked the PR in commit `3e721981ad` following the design feedback.

The full colored header and the global icon-context bridge are gone. The net change is now limited to two theme files: an 8px per-editor accent strip across the full viewport plus five accent color tokens. The neutral header and all light/dark/night/high-contrast behavior remain untouched. PDF forms receive the PDF accent through a specific `body.pdf-form` rule.

I also updated the PR title and description. Production document/PDF builds, standalone LESS compilation, browser-target consistency, and `git diff --check` all pass. Could you please take another look when convenient?

marcoambrosini (approved):
Looks good, but still seems a bit to present, could we reduce the height to 6px?

moodyjmz:
Thanks for turning the rework around so quickly after Marco's feedback, and for keeping it down to two theme files. Two things left: Marco's 6px request, and one functional issue with the border approach.

**The `border-top` on `#viewport` breaks scrollbar hit-testing at any height:** the lower part of the spreadsheet's horizontal scrollbar thumb can't be grabbed, at 8px and at 6px. Fix: offset `#viewport` with `top` and draw the strip as `#viewport::before`, at Marco's 6px.

pr183-scrollbar-crops

<details><summary>Details</summary>

**Cause.** `ScrollObject.getMousePosition` ([sdkjs `common/scroll.js:543-552`](https://github.com/Euro-Office/sdkjs/blob/a14ed31eabff757e9ec9ffc2479139bd6adbc8fc/common/scroll.js#L543-L552)) sums `offsetTop` up the `offsetParent` chain. `offsetTop` doesn't include offsetParent borders, so pointer Y on `ScrollObject` scrollbars reads too low by the border width. Canvas hit-testing is unaffected: the editor canvas uses `getBoundingClientRect`, and a click lands on the correct cell.

**Measured** (Chromium, 1440×900, DPR 1)

| editor | scrollbar | border 8px (as-is) | border 6px | fix, 6px |
|---|---|---|---|---|
| spreadsheet | horizontal | 8px error, 6/13 thumb rows grab | 6px error, 8/13 rows grab | 0, 13/13 rows grab |
| spreadsheet | vertical | 8px | 6px | 0 |
| document | vertical | 8px | 6px | 0 |
| presentation | notes | 9px | 7px | 1px (pre-existing) |

**Repro.** In the spreadsheet editor, press the horizontal scrollbar thumb within a few px of its bottom edge and drag. The sheet doesn't scroll.

**Fix**

```less
// theme.less
@eo-app-accent-height: 6px;

// overrides/header.less
:root body[data-editor-type] #viewport { top: @eo-app-accent-height; }
:root body[data-editor-type] #viewport::before {
  content: ""; position: fixed; top: 0; left: 0; right: 0;
  height: @eo-app-accent-height; background: var(--eo-app-accent);
}
:root body[data-editor-type="document"] { --eo-app-accent: @eo-app-accent-document; }
/* one per editor; pdf-form after document */
```

Scope it to `#viewport`: the forms viewer (`apps/documenteditor/forms/index.html:180`) has `data-editor-type` and imports the theme, but has no `#viewport`.

**Desktop app (not run).** Desktop builds use this theme (`THEME=euro-office`). On win-linux, desktop-apps places its caption mask at y=0 ([`ceditorwindow.cpp:352`](https://github.com/Euro-Office/desktop-apps/blob/303527a8e4f47a7079781fd576816d875801252c/win-linux/src/windows/ceditorwindow.cpp#L352)), so the header will sit lower than the mask by the strip height.

**Validation.** Scrollbar drag and click in the document, spreadsheet and presentation editors, plus the forms viewer rendering unchanged.
</details>

Phillipxh:
Implemented both requested changes in commit `71ceb34871`:

- reduced the accent height from 8px to 6px;
- removed the `border-top` from `#viewport`, offset the viewport with `top`, and draw the strip with the fixed `#viewport::before` pseudo-element;
- moved the per-editor colors to `--eo-app-accent`, with the PDF-form override after the document rule.

Validation completed:
- production webpack builds pass for document, spreadsheet, presentation, PDF, and forms;
- generated CSS parses successfully for all five outputs and contains the 6px offset/pseudo-element with no accent border remaining;
- the forms viewer has no `#viewport`, so the strip/offset remains scoped out there;
- browser-target consistency and `git diff --check` pass.

This environment does not have an interactive browser available, so I could not repeat the physical scrollbar-thumb drag measurement, but the border that caused the offset-parent hit-testing error is no longer present.

@moodyjmz Could you please take another look when convenient?

PR #195, opened 2026-08-13
build: stop tracking the generated mobile index.html

chrip:
`apps/*/mobile/index.html` is webpack output, not source, and it has been tracked
all along. Every build rewrites all four, so they get swept into unrelated pull
requests.

### Why it's output
The framework7 config writes into the source tree and builds the file from a
tracked template:

- `output.path` → `apps/${editor}/mobile` — `vendor/framework7-react/build/webpack.config.js:33`
- `template` → `apps/${editor}/mobile/src/index_dev.html` — `:223`
- stylesheet name → `css/[name].[contenthash].css` via MiniCssExtractPlugin — `:219`

### Why it's inconsistent today
`.gitignore` already ignores every sibling emitted by that same run:

```
apps/**/mobile/dist/**
apps/**/mobile/hot/*.*
apps/**/mobile/css/*.*
```

`index.html` was the one holdout. The only part of it that changes between builds
is the stylesheet `href` — `css/app.<contenthash>.css` — which points into the
`css/` folder ignored on the line directly above. The committed copy therefore
references a file git does not have, and is stale as soon as anyone builds.

### Impact
58 commits have touched these four files, several existing only to re-record
hashes (e.g. `e37812b900 build(mobile): update CSS content hashes after esbuild
migration`).

### The change
One `.gitignore` line beside the existing three, plus `git rm --cached` on the
four files. **−1540 lines. No product code.**

### Verification
Nothing consumes the tracked copy — `build-pipeline` runs the mobile builds
before `deploy-mobile`, and `verify-deploy` asserts the `href` in the *deployed*
`index.html` resolves to a real deployed file.

Checked rather than assumed: deleted all four from disk, ran `make web-apps`,
webpack recreated them with correct hashes, and `git status` then reported
nothing. `git check-ignore -v` confirms the new rule is what matches.


moodyjmz:
**TL;DR:** Approve. The core claim checks out — I independently traced the generation mechanism, ignore glob, and deploy/verify pipeline and they all match the description. One minor nit: the `.gitignore` diff silently flips the whole file from CRLF→LF, which is why it shows as a 102-line change instead of the "one line" the description claims. Cosmetic, but worth fixing so the diff matches its own framing.

<details>
<summary>Full review</summary>

**Verified independently (not just trusting the description):**
- `output.path`, the `index_dev.html` template, and the contenthash-only CSS href all check out against `vendor/framework7-react/build/webpack.config.js:33,219,222-223`. The only volatile line in the tracked files is the `css/app.<hash>.css` href — everything else (`dist/js/app.js`) is a stable filename already covered by the existing `apps/**/mobile/dist/**` ignore rule.
- `build-pipeline.js` runs the mobile webpack builds in Phase 1 before `deploy-mobile` in Phase 2, and `deploy-mobile.js` fails loudly if a mobile `index.html` is missing post-build — so nothing silently ships a stale file.
- `verify-deploy.mjs` asserts the CSS href in the *deployed* `index.html` resolves to a real deployed file — it checks the build output, not the tracked source-tree copy.
- `git check-ignore -v` logic holds: the new glob (`apps/**/mobile/index.html`) matches all four target files and correctly leaves `src/index_dev.html` untouched.
- Diff arithmetic is exact: 508 + 344×3 = 1540 for the four deleted files.
- `make web-apps` (referenced in the description as the verification step) is a real, documented command (`docs/webpack-migration.md`) run via `docker exec eo make web-apps` in the parent `eo` container — confirmed after initially not finding a Makefile in this repo alone.

**Nit — `.gitignore` line-ending flip:**
The base blob is CRLF; the PR's blob is LF-only, so the diff shows 53 additions / 49 deletions for what's actually a single new line plus a comment block. Functionally harmless (git strips `\r` when matching ignore patterns), but it undercuts the PR's own framing ("One `.gitignore` line beside the existing three") and makes the diff harder to review at a glance — a bit ironic for a PR whose whole point is cutting incidental noise. Worth re-saving with CRLF, or just calling out the line-ending change explicitly in the description.

**Informational, not blocking:** open PR #140 ("webpack mirror", parked) touches these same four `index.html` files, and every hunk in it is pure CSS-contenthash churn with zero real content change — a live example of exactly the problem this PR fixes. Once this merges, #140 will need a trivial rebase where those hunks just disappear; not a real conflict.

No issues found with: dev-server-without-build paths, other consumers of the tracked files (checked deploy scripts, CI workflows, docs, build/*.json copy manifests), or the glob accidentally matching anything it shouldn't.

</details>

*(Review assisted by Claude Code — verified independently against source, not taken from the PR description at face value.)*

j-base64:
Reviewed here too as requested

**👍 LOOK GOOD** at simple glance,

with a couple of non-blocking findings surfaced by AI-assisted review. _(double checked myself)_
I'll leave them below at your final appreciation since my eyes are new on the project.


> **Summary**
> No blocking issues found. The core reasoning holds up.
> One cosmetic nit, and two things worth a mention in the description so future-us aren't surprised by them.
> 
> **Nit (non-blocking)**
> - We noticed the same thing @moodyjmz flagged: the `.gitignore` diff flips the whole file from CRLF to LF, which is why it shows as ~100 lines changed instead of the one line the description mentions. Harmless functionally (git ignores `\r` when matching patterns), just makes the diff and future `git blame` on that file noisier than it needs to be. Re-saving with CRLF would make the diff match its own framing 
> 
> **Worth a mention in the description (non-blocking)**
> - `deploy-mobile.js` already fails loudly (exit 1) if a mobile `index.html` is missing post-build — that logic isn't new here, but its behavior in practice does shift: before, a skipped/killed webpack step would silently deploy the stale-but-tracked file; after this change, the same scenario now hard-fails instead. We think that's actually an improvement (fail loud beats silent stale deploy), but since it's not mentioned in the PR body, we wanted to flag it so it isn't mistaken for a new bug the first time someone hits it.
> - Untracking the generated file also means the final rendered HTML (post `HtmlWebpackPlugin` injection) no longer shows up in future diffs — the template and webpack config stay reviewable, but the actual output doesn't. Not something this PR needs to solve, just a tradeoff worth naming out loud since it's a small step down in "what a reviewer can see without rebuilding locally."
> 
> _Assisted by Claude Sonnet 5_

🙌

j-base64:
Looks like a commit fixing the CRLF landed while I was writing the previous comment 🫣👍.
I'll let you crosscheck the other notes.

chrip:
@moodyjmz @j-base64 I fixed the line endings. Can one of you approve it?

moodyjmz (approved):
**TL;DR:** Re-verified independently — the CRLF fix mentioned above never landed, because there was never anything to fix. Approving regardless; the core change is correct.

<details>
<summary>Why</summary>

Checked commit history and the raw `.gitignore` patch directly rather than trusting the diff summary:

- `gh api repos/Euro-Office/web-apps/pulls/195/commits` — one commit, `21f4eda1d06265ef400adf0f8793c502039a3d1d`, no amends.
- One `head_ref_force_pushed` event, timestamped the same minute as that commit — the initial push, not a later rewrite.
- The `.gitignore` patch has been `+4/-0` the entire time:

  ```
  +# Emitted by the same webpack run as dist/ and css/ above, from the tracked
  +# template apps/*/mobile/src/index_dev.html. Its only volatile content is the
  +# contenthash-named css/ href, which is ignored on the line above.
  +apps/**/mobile/index.html
  ```

No CRLF churn, before or after. The "fix" neither of us actually diffed landed — good reminder to check the diff before confirming an AI-flagged nit is resolved, not just the follow-up comment saying it is.

Rest of the PR holds up: `output.path`/template/CSS-hash claims match `vendor/framework7-react/build/webpack.config.js:33,219,221-223` exactly, `build-pipeline.js` always runs the mobile builds before `deploy-mobile`, `deploy-mobile.js` fails loudly (not silently) on a missing file, and `verify-deploy.mjs` checks the deployed output only. Approving.

</details>

Issue #65, opened 2026-05-26
Bug: Missing icons for buttons in the right panel


rikled:
The buttons in the right panel are missing their icons:

<img width="1917" height="1035" alt="Image" src="https://github.com/user-attachments/assets/a775ff45-3315-4582-bc6d-a3f0e2e4643d" />

moodyjmz:
are added a different way to other icons, and were removed with the png sweep. I will use svgs

moodyjmz:
<img width="45" height="333" alt="Image" src="https://github.com/user-attachments/assets/b0f2b309-5fb5-4884-81e3-ada709192e1a" />

<img width="466" height="82" alt="Image" src="https://github.com/user-attachments/assets/e2fb6df4-a0a8-4a55-a932-c7e3656fd512" />

PR #81, opened 2026-06-09
Translate Russian comments to English

jvitkauskas:
This compliments effort in https://github.com/Euro-Office/core/pull/87

Please note that Cyrillic "С" letter (homoglyph) was replaced with latin letter C in "Сommenting", variable name effectСolors and variable name _СhartSettingsDlg_view

PR #197, opened 2026-08-15
fix: convert static <i> icons of el-bound buttons to SVG sprite references

### LINKED ISSUE #196: Status bar icons invisible in desktop builds (add worksheet, sheet list, zoom +/-) — SVG icon migration misses el-bound buttons
**Symptom.** In self-built DesktopEditors (macOS arm64 and Windows x64, current `main`), several spreadsheet status bar buttons render blank while remaining functional (hover tooltips still appear): add worksheet, sheet list, tab prev/next, and zoom +/-.

**Root cause.** The SVG sprite migration (`templateBtnIcon` in `apps/common/main/lib/component/Button.js`) emits `<svg><use href="#btn-x">` only when a button renders its own template. Buttons bound to pre-existing markup via the `el` option — e.g. everything in `apps/spreadsheeteditor/main/app/template/StatusBar.template` — never run the template, so their static `<i class="icon toolbar__icon btn-*">` markup survives. The CSS sprite classes that markup relied on no longer exist in the built `app.css`, leaving functional but invisible buttons. Any el-bound button with static `<i>` icons in any editor is affected.

**Fix.** The linked PR converts those static icons in `render()` the same way the template does (same `/btn-[^\s]+/` extraction, fragment-only `<use>` references). Verified on macOS and Windows desktop builds: all status bar icons render again; add/list/zoom behaviors unchanged.

lif9e1:
Fixes #196.

Buttons bound to existing markup via `el` skip `templateBtnIcon`, so their static `<i class="icon toolbar__icon btn-*">` elements were left relying on css sprite classes that the SVG migration removed — functional but invisible buttons (spreadsheet status bar: add worksheet, sheet list, tab prev/next, zoom +/-).

This adds the missing half of the migration in `Button.render()`: static icons are converted with the same `/btn-[^\s]+/` extraction and fragment-only `<use>` references the template uses. 11 lines, no behavior change for template-rendered buttons.

Tested: macOS arm64 + Windows x64 desktop builds, spreadsheet/document/presentation/pdf editors rebuilt with the patch; icons render in all themes tried (light/dark).

### DIFF
--- a/apps/common/main/lib/component/Button.js
+++ b/apps/common/main/lib/component/Button.js
@@ -469,6 +469,17 @@ define([
                     parentEl.html(me.cmpEl);
                     me.$icon = me.$el.find('.icon');
                 }
+            } else if (!me.rendered) {
+                // SVG sprite approach (see templateBtnIcon): buttons bound to an
+                // existing element via `el` never run the template, so their static
+                // <i class="icon toolbar__icon btn-*"> markup survives — but the css
+                // sprite classes it relied on are gone. Convert it here the same way
+                // the template does, keeping fragment-only references.
+                me.cmpEl.find('i.icon.toolbar__icon').each(function () {
+                    var iconMatch = /btn-[^\s]+/.exec(this.className);
+                    if (iconMatch)
+                        $(this).replaceWith('<svg class="icon uni-scale"><use href="#' + iconMatch[0] + '"></use></svg>');
+                });
             }
 
             if (!me.rendered) {


moodyjmz:
Not marking this as changes-requested, because the bug you reported is real — it's just not in this repo. Thanks for chasing it as far as you did.

**The patch is a no-op on this branch.** `apps/spreadsheeteditor/main/app/template/StatusBar.template` at your head commit has zero `<i class=` and six `<use href="#btn-...">`, on lines 4/5/8/10/15/21 — exactly the tab prev/next, add worksheet, sheet list and zoom ± buttons the description lists. They were converted in `de4653f035` (2026-05-27, "Fixes #65"), which is an ancestor of `fix/el-bound-buttons-svg-icons`; your branch point `73bdfbb0b7` already has it.

**What you actually hit was our stale submodule pin, not your build.** DesktopEditors pins `web-apps` at `c6368d5eaa` ("remove obsolete DPI-scaled PNG toolbar icon spritesheets", 2026-05-13) — the commit that deleted the sprites, two weeks *before* the templates were converted. It's 232 commits behind, and at that pin the broken `<i class="icon toolbar__icon btn-*">` markup is still present in every editor. So you built DesktopEditors' `main`, saw genuinely broken icons, and diagnosed against web-apps `main` where it had been fixed for three months. That trap is on us. Filed as Euro-Office/DesktopEditors#81.

**Separately, the selector misses the cases that *are* still broken here.** It requires `toolbar__icon`; every surviving el-bound button with static sprite markup uses `menu__icon`:

- `apps/common/main/lib/component/UpDownPicker.js:52,54` — `<i class="icon menu__icon <%= options.iconUpCls %>">`, defaults `btn-zoomup`/`btn-zoomdown` (`:44-45`), bound at `:64,70` via `new Common.UI.Button({el: ...})`. This is common code, live in `common/main/lib/view/Draw.js:221`, `pdfeditor/.../Toolbar.js:1694,1805`, `spreadsheeteditor/.../Toolbar.js:2502`. These are the zoom buttons that really are blank on current main.
- `apps/documenteditor/main/app/view/Links.js:334-368`, bound at `:391-409` — the footnote/endnote goto arrows.

`i.icon[class*="btn-"]` would catch both.

**And if you do rework it**, don't add a second conversion path — #203 landed `Common.UI.menuItemIconMarkup` in `MenuItem.js` and it preserves the original class list. This version hardcodes `class="icon uni-scale"` and drops `icon-rtl`, which `applyScaling` (`Button.js:1025`) deliberately carries through.

Suggest closing this one. If you want to take the `menu__icon` cases as a follow-up PR that'd be genuinely useful — otherwise I'll pick it up alongside #81.

moodyjmz:
Closing as is a no-op, thanks to @lif9e1 for finding it though

PR #202, opened 2026-08-18
PostMessage: add mentions via WOPI

rawe0:
Hi

I work at CERN in the CERNBox team and we're about to start a EuroOffice pilot in CERNBox and mentions came up as a gap: opened through WOPI, the @/+ button never appears in the comment pane, because the host has no way to answer the editor's user-list request. The callbacks already exist in the editor - they're just not wired up in the WOPI page, which registers onRequestSharingSettings, onRequestHistory and friends but stops short of the mention set.

All changes are in apps/api/wopi/editor-wopi.ejs; nothing in the editors themselves changes. It adds three exchanges between the editor and the host:

- when the user types @, the editor asks the host who it can suggest, and the host answers with a list of names and emails
- when the comment is posted, the editor hands the host the message, the mentioned addresses and a link anchor for that comment, so the host can send the notification
- "Get Link" - on a bookmark in Documents, on a cell range in Spreadsheets - asks the host for a url pointing at that spot

Hosts opt in with a new CheckFileInfo property, UserMentionPostMessage, the same way FileSharingPostMessage switches on the sharing button today - hosts that don't set it see exactly today's behaviour. Mentions are also hidden from users who can't comment. "Get Link" isn't, since that's useful to a read-only reader too.

This has been tested with our [wopiserver](https://github.com/cs3org/wopiserver/).

### DIFF
--- a/apps/api/wopi/editor-wopi.ejs
+++ b/apps/api/wopi/editor-wopi.ejs
@@ -129,6 +129,23 @@ div {
                     "images": [{ "url": data.Values.url }]
                 });
             },
+            'Action_SetUsers': function (data) {
+                //response to UI_RequestUsers; "c" is echoed back by the host so the editor can
+                //route the list to the request that asked for it. Each user needs both "email"
+                //and "name" or Common.UI.ExternalUsers drops it from the list; "hasAccess"
+                //sorts users who can already open the file above a separator, "image" is the
+                //optional avatar. isPaginated must be present to keep the editor in
+                //server-side-search mode (undefined switches it to "fetch once, filter locally").
+                data && data.Values && docEditor.setUsers({
+                    "c": data.Values.c,
+                    "users": data.Values.users,
+                    "isPaginated": data.Values.isPaginated
+                });
+            },
+            'Action_SetActionLink': function (data) {
+                //response to UI_MakeActionLink; setActionLink takes the url as a bare string
+                data && data.Values && docEditor.setActionLink(data.Values.url);
+            },
             'Host_PostmessageReady': function (data) {
                 innerAlert('Host_PostmessageReady');
             }
@@ -248,6 +265,42 @@ div {
         var onRequestInsertImage = function (event) {
             insertImageType = event.data.c;
             _postMessage('UI_InsertGraphic', {});
+        };
+
+        var onRequestUsers = function (event) {
+            //see Common.Gateway.requestUsers: "c" is the operation type (mention/protect/info),
+            //"id" carries the user ids to resolve for c==="info", and from/count/search describe
+            //the mention autocomplete query. Forwarding search/from/count lets the host search
+            //server-side; it must then answer with isPaginated set, otherwise
+            //Common.UI.ExternalUsers caches the first response and never asks again.
+            _postMessage('UI_RequestUsers', {
+                "c": event.data.c,
+                "id": event.data.id,
+                "from": event.data.from,
+                "count": event.data.count,
+                "search": event.data.search
+            });
+        };
+
+        var onRequestSendNotify = function (event) {
+            //event.data is Comments.js' pickEMail payload: the emails are parsed out of the
+            //comment text, so they are exactly the ones the host handed over in Action_SetUsers
+            _postMessage('UI_SendNotify', {
+                "actionLink": event.data.actionLink,
+                "actionId": event.data.actionId,
+                "message": event.data.message,
+                "emails": event.data.emails
+            });
+        };
+
+        var onMakeActionLink = function (event) {
+            //event.data is {action: {type: "bookmark"|"internallink", data: <bookmark name|sheet!range>}};
+            //the host turns it into a document url carrying that anchor, and answers
+            //Action_SetActionLink. Comment links do not come through here - Comments.js builds those
+            //itself and passes them along with UI_SendNotify.
+            _postMessage('UI_MakeActionLink', {
+                "config": event.data
+            });
         };
 
         var onError = function (event) {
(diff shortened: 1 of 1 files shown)


moodyjmz:
Thanks for this — nicely researched, and the plumbing all checks out end-to-end against `Gateway.js`/`api.js`/`ExternalUsers.js`.

One gating issue worth fixing before merge: `onRequestUsers` is scoped to `permissionsComment && fileInfo.UserMentionPostMessage` at [editor-wopi.ejs:466](https://github.com/Euro-Office/web-apps/blob/1d959361a796e0e60e3d5b2fefa6cd1b80695158/apps/api/wopi/editor-wopi.ejs#L466), but `onRequestUsers` isn't only used for mentions — it also handles `c === 'info'` requests, which resolve user avatar images and are fired unconditionally regardless of permissions (e.g. `documenteditor/main/app/controller/Main.js:1823` calls `ExternalUsers.get('info', appOptions.user.id)` on load for every user). `ExternalUsers.js`'s `_get()` has no permission check before calling `Common.Gateway.requestUsers('info', ids)`.

With the current gate, a host that sets `UserMentionPostMessage` will get working avatars for editors/commenters, but view-only users will silently get none — the `info` request just has nowhere to go once the handler is `undefined`.

Suggested fix: gate `onRequestUsers` on `UserMentionPostMessage` alone, and keep `permissionsComment` only on `onRequestSendNotify`. That still keeps the mention UI itself away from viewers — the @/+ affordance is driven by `canRequestSendNotify`, and the autocomplete lives inside a comment textbox a viewer never has open — so this shouldn't change any user-facing mention behavior, just fix the avatar side effect.

rawe0:
> Thanks for this — nicely researched, and the plumbing all checks out end-to-end against `Gateway.js`/`api.js`/`ExternalUsers.js`.
> 
> One gating issue worth fixing before merge: `onRequestUsers` is scoped to `permissionsComment && fileInfo.UserMentionPostMessage` at [editor-wopi.ejs:466](https://github.com/Euro-Office/web-apps/blob/1d959361a796e0e60e3d5b2fefa6cd1b80695158/apps/api/wopi/editor-wopi.ejs#L466), but `onRequestUsers` isn't only used for mentions — it also handles `c === 'info'` requests, which resolve user avatar images and are fired unconditionally regardless of permissions (e.g. `documenteditor/main/app/controller/Main.js:1823` calls `ExternalUsers.get('info', appOptions.user.id)` on load for every user). `ExternalUsers.js`'s `_get()` has no permission check before calling `Common.Gateway.requestUsers('info', ids)`.
> 
> With the current gate, a host that sets `UserMentionPostMessage` will get working avatars for editors/commenters, but view-only users will silently get none — the `info` request just has nowhere to go once the handler is `undefined`.
> 
> Suggested fix: gate `onRequestUsers` on `UserMentionPostMessage` alone, and keep `permissionsComment` only on `onRequestSendNotify`. That still keeps the mention UI itself away from viewers — the @/+ affordance is driven by `canRequestSendNotify`, and the autocomplete lives inside a comment textbox a viewer never has open — so this shouldn't change any user-facing mention behavior, just fix the avatar side effect.

Hi,

Yep this makes sense, I've updated the PR!

rawe0:
Hi @moodyjmz any update on this? I've changed the code as you suggested.

moodyjmz:
> Hi @moodyjmz any update on this? I've changed the code as you suggested.

Hi sorry, missed this, I'll take a look shortly and get back to you

apps/api/wopi/editor-wopi.ejs
```diff
@@ -250,6 +267,42 @@ div {
             _postMessage('UI_InsertGraphic', {});
         };
 
+        var onRequestUsers = function (event) {
+            //see Common.Gateway.requestUsers: "c" is the operation type (mention/protect/info),
+            //"id" carries the user ids to resolve for c==="info", and from/count/search describe
+            //the mention autocomplete query. Forwarding search/from/count lets the host search
+            //server-side; it must then answer with isPaginated set, otherwise
+            //Common.UI.ExternalUsers caches the first response and never asks again.
```
moodyjmz, line 275:
```suggestion
            //Common.UI.ExternalUsers caches the first response and never asks again.
            //this handler is gated only on UserMentionPostMessage, not on the requesting
            //user's comment permission - c==="info" (avatars) is safe to answer for anyone,
            //but c==="mention" is a searchable name+email directory and per-user
            //comment-right checking is unenforced here.
```

apps/api/wopi/editor-wopi.ejs
```diff
@@ -250,6 +267,42 @@ div {
...
+                "id": event.data.id,
+                "from": event.data.from,
+                "count": event.data.count,
+                "search": event.data.search
+            });
+        };
+
+        var onRequestSendNotify = function (event) {
+            //event.data is Comments.js' pickEMail payload: the emails are parsed out of the
+            //comment text, so they are exactly the ones the host handed over in Action_SetUsers
```
moodyjmz:
```suggestion
            //event.data is Comments.js' pickEMail payload, extracted via regex from the raw
            //comment text - NOT validated against the list Action_SetUsers returned. a
            //commenter can type any address here; hosts must validate "emails" against their
            //own ACLs before sending, same as any other client-supplied string in this file.
```

moodyjmz (approved):
Approving — fix looks correct, mention/notify/action-link wiring round-trips cleanly. Left two inline suggestions (on `onRequestUsers` and `onRequestSendNotify`) with comment notes for whoever implements the WOPI host side — worth committing since they document assumptions a host needs to get right, but not blockers.

One question: is bundling mentions + notify + action-link under one `UserMentionPostMessage` flag intentional? A host that only wants avatars/mentions also silently gets "copy link to bookmark." Not a blocker either way.

@rawe0

rawe0:
> Approving — fix looks correct, mention/notify/action-link wiring round-trips cleanly. Left two inline suggestions (on `onRequestUsers` and `onRequestSendNotify`) with comment notes for whoever implements the WOPI host side — worth committing since they document assumptions a host needs to get right, but not blockers.
> 
> One question: is bundling mentions + notify + action-link under one `UserMentionPostMessage` flag intentional? A host that only wants avatars/mentions also silently gets "copy link to bookmark." Not a blocker either way.
> 
> @rawe0

This was mostly for convenience but we can change that if you would like?

Sorry for the late reply as well, I've been offline for the last two weeks. 

@moodyjmz

moodyjmz:
@rawe0 Fine with leaving as is. Sorry for the late reply, again. 😰

rawe0:
> @rawe0 Fine with leaving as is. Sorry for the late reply, again. 😰

Okay, thanks! I'm not able to merge since I don't have write rights

PR #203, opened 2026-08-18
fix(menus): draw menu icons from the symbol sheet, not the dead CSS sprite

### LINKED ISSUE #313: Dropdown menu items still render icons the old way — every menu icon is blank (all editors)
## Summary

Toolbar icons render fine here, which I assume is the result of the #199 fix. Menu
icons do not: every dropdown menu item that declares an icon shows an empty 20×20
box where the icon should be. Captions, shortcuts and behaviour are all correct.

`Common.UI.Button` was migrated to the SVG symbol sprite, but `Common.UI.MenuItem`
was not. It still emits the old CSS-sprite `<span>`, and the PNG sprites that
mechanism needs (`iconssmall*.png`, `iconsbig*.png`, `iconshuge*.png`) are no longer
generated or shipped, so nothing paints.

This is unchanged in v9.3.3: `apps/common/main/lib/component/Menu.js` is identical
between the web-apps submodule of v9.3.1 (`7c69898`) and of v9.3.3 (`91f0f97`).

## Where

`apps/common/main/lib/component/Menu.js:846-848`, in `MenuItem.itemTemplate`:

```js
'<% if (typeof iconCls !== "undefined") { %>',
    '<span class="menu-item-icon <%= iconCls %>"></span>',
'<% } %>',
```

`apps/common/main/lib/component/Button.js:176-187`, which was migrated:

```js
// SVG sprite approach - uses <svg><use href="#id"> for dark mode support
// Sprite is injected into DOM via svg-injector, so we use fragment-only references
var templateBtnIcon =
        '<% if ( iconImg ) { %>' +
            '<img src="<%= iconImg %>">' +
        '<% } else { %>' +
            '<% var iconMatch = /btn-[^\\s]+/.exec(iconCls); ' +
            'if (iconMatch) {' +
                'print(\'<svg class=\"icon uni-scale\"><use href=\"#\' + iconMatch[0] + \'\"></use></svg>\');' +
            '} else ' +
                'print(\'<i class=\"icon \' + iconCls + \'\">&nbsp;</i>\'); %>' +
        '<% } %>';
```

The only rule matching the markup `MenuItem` emits is geometry — there is no image
anywhere in the chain:

```css
.dropdown-menu li .menu-item-icon {
  display: inline-block; float: left;
  width: var(--x-small-btn-icon-size, 20px);
  height: var(--x-small-btn-icon-size, 20px);
  margin: calc((16px - var(--x-small-btn-icon-size, 20px)) / 2) 4px 0 -16px;
  background-repeat: no-repeat;
  opacity: var(--component-normal-icon-opacity, 0.8);
}
```

The `background-image` / `background-position` for each icon class used to come from
`apps/<editor>/main/resources/less/sprites/*.less`. Those files no longer exist and
their `@import`s are commented out in every `app.less`. Grepping the built `app.css`
for `.btn-*` selectors returns 55 hits, all of them component classes
(`.btn-default`, `.btn-danger`, `.btn-toolbar`, …) — no icon classes at all.

## Impact

Menu items declaring an icon, counted on v9.3.3 sources:

| editor | items |
| --- | --- |
| documenteditor | 315 |
| presentationeditor | 298 |
| pdfeditor | 266 |
| spreadsheeteditor | 194 |
| visioeditor | 19 |
| **total** | **1092** |

Where it shows most, in the document editor:

- `app/view/DocumentHolderExt.js` — 273 (right-click context menu and its submenus)
- `app/view/FileMenu.js` — 21 (File menu)
- `app/view/Toolbar.js` — 13 (toolbar dropdowns)

## The sprite already contains what is needed

The 91 distinct icons requested by the document editor's menus are all present as
`<symbol id="btn-…">` in the shipped
`apps/documenteditor/main/resources/img/toolbar/icons.svg`. So no artwork is missing
— only the template needs to emit `<use>`.

Aligning `MenuItem.itemTemplate` with the button approach should be enough:

```js
'<% if (typeof iconCls !== "undefined") { %>',
    '<% var iconMatch = /btn-[^\\s]+/.exec(iconCls); %>',
    '<% if (iconMatch) { %>',
        '<svg class="menu-item-icon icon"><use href="#<%= iconMatch[0] %>"></use></svg>',
    '<% } else { %>',
        '<span class="menu-item-icon <%= iconCls %>"></span>',
    '<% } %>',
'<% } %>',
```

`.dropdown-menu li .menu-item-icon` then needs `fill: currentColor` so the symbols
follow the theme colour, the way `.svg-icon` does in `icons-svg.less`.

One related spot to check while fixing: `Menu.js:1054`, in `setChecked()`, still
clears the icon with `itemEl.css('background-image', 'none')`. That assumes the icon
is a CSS background, so it will need the same treatment once menu icons become
`<svg><use>`.

## Smaller findings in the same area

1. `apps/common/main/lib/util/themeinit.js:107` requests three sprite files the build
   never produces — `iconssmall@2.5x.svg`, `iconsbig@2.5x.svg`, `iconshuge@2.5x.svg`.
   `build/scripts/deploy-sprites.js` only emits `toolbar/icons.svg` per editor plus
   `doc-formats/formats@2.5x.svg`. That is three 404s on every editor load, and any
   path relying on `apply_icons_from_url()` to re-inject icons (theme switch, custom
   themes with their own `icons.basepath`) gets nothing back.

2. `apps/common/main/resources/less/colors-table.less:205-224` still defines 15
   `--sprite-button-{small,big,huge}-{100,125,150,175,200}` variables pointing at
   `iconssmall*.png` / `iconsbig*.png` / `iconshuge*.png`, none of which exist any
   more. Nothing consumes them in the built CSS, so they are harmless, but they are
   what sends you chasing missing PNGs when tracing this.

3. `btn-find-redacted` is referenced by the document editor but has no symbol in
   `icons.svg`.

4. `apps/common/main/resources/img/plugin/icon_add_on_default@1.5x.png` → 404, while
   the 1x and 2x variants ship. Default plugin icon is blank on 1.5x displays.

5. `apps/common/main/resources/img/doc-formats/large/hwp.svg` and `hwpx.svg` → 404,
   both referenced by the format-list CSS.

## Environment

- `ghcr.io/euro-office/documentserver`, amd64, served resources report 9.3.3
- Behind Traefik, TLS terminated at the proxy
- Integrated through a small self-hosted file manager (JWT + callback), not Nextcloud
- All 404s above verified over HTTP against the running instance; code references
  checked against the v9.3.3 sources (web-apps `91f0f97`)

Image

Image

Image

chrip:
Every dropdown menu item that declares an icon showed an empty box, in all editors. Common.UI.Button was migrated to the SVG symbol sheet; Common.UI.MenuItem was not, and still emitted

    <span class="menu-item-icon btn-...">

whose only paint came from apps/<editor>/main/resources/less/sprites -- files that no longer exist, positioning PNGs the build no longer produces. The artwork was never missing: every symbol the menus ask for is already in the shipped icons.svg.

Common.UI.menuItemIconMarkup() now returns the same <svg class="menu-item-icon uni-scale"><use href="#btn-..."></svg> that Button prints, and every menu template calls it. uni-scale is what tells bigscaling.less the svg is valid at any device pixel ratio.

A class that is not a sprite name still gets a span, because some of them are not icons at all -- ColorButton's "Automatic" swatch and the border-colour pickers use it as a coloured block.

Six templates outside MenuItem hardcoded the same span and are switched over too: the editing-mode menu in Header, the thumbnail-size sliders in Writer and the PDF editor, and the form zoom items in FormsTab and FormSettings.

setIconCls looked for `use.zoom-int`, the class applyScaling gives the svg it injects at ratio > 2. The template's own <use> carries no class, so it now matches either.

#file-menu-panel hid svg.menu-item-icon and showed the span, which was right when the span was the icon and is backwards now.

Measured in Writer at devicePixelRatio 1, before -> after: 92 empty spans -> 92 rendered svgs, 3 spans left (the colour swatches). Verified in Writer, Spreadsheet and Presentation: context menus, the File menu, the editing-mode menu and the colour menus.

Fixes Euro-Office/DocumentServer#313

### DIFF
--- a/apps/common/main/lib/component/Menu.js
+++ b/apps/common/main/lib/component/Menu.js
@@ -844,7 +844,7 @@ define([
                                                                     '<% if(typeof stopPropagation !== "undefined") { %> data-stopPropagation="true" <% } %>',
                                                                     'class="<% if (checked) { %> checked <% } %>" >',
                                                                     '<% if (typeof iconCls !== "undefined") { %>',
-                                                                        '<span class="menu-item-icon <%= iconCls %>"></span>',
+                                                                        '<%= Common.UI.menuItemIconMarkup(iconCls) %>',
                                                                     '<% } %>',
                                                                 '<%= caption %>',
                                                                 '</a>'
--- a/apps/common/main/lib/component/MenuItem.js
+++ b/apps/common/main/lib/component/MenuItem.js
@@ -78,6 +78,34 @@ define([
     'common/main/lib/component/ToggleManager'
 ], function () {
     'use strict';
+    /**
+     * Icon markup for a menu item.
+     *
+     * Sprite icons are drawn from the symbol sheet, the way Common.UI.Button
+     * does it. The <span> this used to emit was painted by a CSS background
+     * from apps/<editor>/main/resources/less/sprites -- files that no longer
+     * exist, generating PNG sprites the build no longer produces -- so every
+     * menu icon in every editor came out as an empty box. The artwork was
+     * never missing: the symbols are all in the shipped icons.svg.
+     *
+     * A class that is not a sprite name still gets the span: ColorButton's
+     * swatches and the border-colour pickers use one as a coloured block, not
+     * as an icon.
+     *
+     * uni-scale is what tells bigscaling.less this svg is valid at any device
+     * pixel ratio, which is also how Button marks its own icons.
+     *
+     * @param {String} iconCls the item's iconCls
+     * @return {String} markup for the icon element
+     */
+    Common.UI.menuItemIconMarkup = function (iconCls) {
+        var cls = iconCls || '',
+            match = /btn-[^\s]+/.exec(cls);
+        return match
+            ? '<svg class="menu-item-icon uni-scale ' + cls + '"><use href="#' + match[0] + '"></use></svg>'
+            : '<span class="menu-item-icon ' + cls + '"></span>';
+    };
+
 
     Common.UI.MenuItem = Common.UI.BaseView.extend({
         options : {
@@ -108,7 +136,7 @@ define([
         template: _.template([
             '<% if (header) { %><span class="menu-item-header"><%- header %></span><% } %><% if (caption) { %><a id="<%= id %>" class="menu-item" <% if (_.isEmpty(iconCls)) { %> data-no-icon <% } %> style="<%= style %>" <% if(options.canFocused) { %> tabindex="-1" type="menuitem" <% }; if(!_.isUndefined(options.stopPropagation)) { %> data-stopPropagation="true" <% }; if(!_.isUndefined(options.dataHint)) { %> data-hint="<%= options.dataHint %>" <% }; if(!_.isUndefined(options.dataHintDirection)) { %> data-hint-direction="<%= options.dataHintDirection %>" <% }; if(!_.isUndefined(options.dataHintOffset)) { %> data-hint-offset="<%= options.dataHintOffset %>" <% }; if(options.dataHintTitle) { %> data-hint-title="<%= options.dataHintTitle %>" <% }; %> >',
                 '<% if (!_.isEmpty(iconCls)) { %>',
-                    '<span class="menu-item-icon <%= iconCls %>"></span>',
+                    '<%= Common.UI.menuItemIconMarkup(iconCls) %>',
                 '<% } else if (!_.isEmpty(iconImg)) { %>',
                     '<img src="<%= iconImg %>" class="menu-item-icon">',
                 '<% } %>',
@@ -279,7 +307,9 @@ define([
                 var firstChild = this.cmpEl.children(':first');
                 if (firstChild) {
                     firstChild.find('.menu-item-icon').removeClass(this.iconCls).addClass(iconCls);
-                    var svgIcon = firstChild.find('use.zoom-int');
+                    // The template's own <use> carries no class; the one
+                    // applyScaling injects is .zoom-int. Match either.
+                    var svgIcon = firstChild.find('.menu-item-icon use, use.zoom-int');
                     if (svgIcon.length) {
                         var re_icon_name = /btn-[^\s]+/.exec(iconCls),
                             icon_name = re_icon_name ? re_icon_name[0] : "null";
--- a/apps/common/main/lib/view/Header.js
+++ b/apps/common/main/lib/view/Header.js
@@ -597,7 +597,7 @@ define([
 
             var menuTemplate = _.template('<a id="<%= id %>" tabindex="-1" type="menuitem" class="menu-item"><div>' +
                                             '<% if (!_.isEmpty(iconCls)) { %>' +
-                                                '<span class="menu-item-icon <%= iconCls %>"></span>' +
+                                                '<%= Common.UI.menuItemIconMarkup(iconCls) %>' +
                                             '<% } %>' +
                                             '<b><%= caption %></b></div>' +
                                             '<% if (options.description !== null) { %><label class="margin-left-10 description"><%= options.description %></label>' +
--- a/apps/common/main/resources/less/common.less
+++ b/apps/common/main/resources/less/common.less
@@ -240,8 +240,11 @@ label {
             border-left: @scaled-one-px-value solid @border-toolbar;
         }
 
+        // Inverted with the move to the symbol sheet: the icon is the <svg>
+        // now, and the <span> only survives for classes that are not sprite
+        // names. Hiding the svg here left the File menu without icons.
         svg.menu-item-icon {
-            display: none;
+            display: block;
         }
 
         span.menu-item-icon {
--- a/apps/documenteditor/main/app/view/FormSettings.js
+++ b/apps/documenteditor/main/app/view/FormSettings.js
@@ -634,7 +634,7 @@ define([
                         '<li id="<%= item.id %>" data-value="<%= Common.Utils.String.htmlEncode(item.value) %>"<% if (item.value === 0) { %> class="border-top"<% } %>>',
                             '<% if (item.value === 0) { %>',
                                 '<a tabindex="-1" type="menuitem" style="display: block; padding: ' + (Common.UI.isRTL() ? '5px 24px 5px 20px' : '5px 20px 5px 24px') + ';">',
-                                    '<span class="menu-item-icon menu__icon btn-zoomup"></span>',
+                                    '<%= Common.UI.menuItemIconMarkup("menu__icon btn-zoomup") %>',
                                     '<%= Common.Utils.String.htmlEncode(item.displayValue) %>',
                                 '</a>',
                             '<% } else { %>',
--- a/apps/documenteditor/main/app/view/FormsTab.js
+++ b/apps/documenteditor/main/app/view/FormsTab.js
@@ -570,7 +570,7 @@ define([
                             '<li id="<%= item.id %>" data-value="<%= Common.Utils.String.htmlEncode(item.value) %>"<% if (item.value === 0) { %> class="border-top"<% } %>>',
                                 '<% if (item.value === 0) { %>',
                                 '<a tabindex="-1" type="menuitem" style="display: block; padding: ' + (Common.UI.isRTL() ? '5px 24px 5px 20px' : '5px 20px 5px 24px') + ';">',
-                                    '<span class="menu-item-icon menu__icon btn-zoomup"></span>',
+                                    '<%= Common.UI.menuItemIconMarkup("menu__icon btn-zoomup") %>',
                                     '<%= Common.Utils.String.htmlEncode(item.displayValue) %>',
                                 '</a>',
                                 '<% } else { %>',
@@ -729,7 +729,7 @@ define([
                         // }
                         var menuTemplate = _.template('<a id="<%= id %>" tabindex="-1" type="menuitem" class="menu-item">'+
                             '<% if (!_.isEmpty(iconCls)) { %>'+
-                            '<span class="menu-item-icon <%= iconCls %>"></span>'+
+                            '<%= Common.UI.menuItemIconMarkup(iconCls) %>'+
                             '<% } %>'+
                             '<div><%= caption %></div>' +
                             '<% if (options.description !== null) { %><label style="cursor: pointer;white-space: normal;"><%= options.description %></label>' +
--- a/apps/documenteditor/main/app/view/PageThumbnails.js
+++ b/apps/documenteditor/main/app/view/PageThumbnails.js
@@ -78,9 +78,9 @@ define([
                             '<% } %>', '>',
                             '<label>' + this.textThumbnailsSize + '</label>',
                             '<div class="thumbnails-sld-box">',
-                                '<span class="menu-item-icon menu__icon btn-thumbnail-small"></span>',
+                                '<%= Common.UI.menuItemIconMarkup("menu__icon btn-thumbnail-small") %>',
                                 '<div id="sld-thumbnails-size"></div>',
-                                '<span class="menu-item-icon menu__icon btn-thumbnail-big"></span>',
+                                '<%= Common.UI.menuItemIconMarkup("menu__icon btn-thumbnail-big") %>',
                             '</div>',
                         '</div>'
                     ].join('')),
--- a/apps/pdfeditor/main/app/view/PageThumbnails.js
+++ b/apps/pdfeditor/main/app/view/PageThumbnails.js
@@ -78,9 +78,9 @@ define([
                             '<% } %>', '>',
                             '<label>' + this.textThumbnailsSize + '</label>',
                             '<div class="thumbnails-sld-box">',
-                                '<span class="menu-item-icon menu__icon btn-thumbnail-small"></span>',
+                                '<%= Common.UI.menuItemIconMarkup("menu__icon btn-thumbnail-small") %>',
                                 '<div id="sld-thumbnails-size"></div>',
-                                '<span class="menu-item-icon menu__icon btn-thumbnail-big"></span>',
+                                '<%= Common.UI.menuItemIconMarkup("menu__icon btn-thumbnail-big") %>',
                             '</div>',
                         '</div>'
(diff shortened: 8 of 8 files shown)


MonaAghili (commented):
## Findings

### [MEDIUM] `apps/common/main/lib/component/MenuItem.js:305-321` — `setIconCls()` can't change icon *kind*, only its classes

`setIconCls()` never replaces the icon element, it only does `find('.menu-item-icon').removeClass(this.iconCls).addClass(iconCls)`. Since the migration made the element type (`<svg>` vs `<span>`) depend on whether `iconCls` contains a `btn-*` token, any call that flips a menu item from a sprite icon to a swatch class (or vice versa) leaves the wrong tag in the DOM:

- btn-* → swatch: the `<svg><use></use></svg>` stays, just gets the swatch class added; `svgIcon.length` is still truthy (the old `<use>` is still there) so the code does `svgIcon.attr('href', '#null')` (no `btn-*` match) — a broken, empty `<svg>` where a colored swatch was expected.
- swatch → btn-*: the `<span>` just gets the new `btn-*` class added; there's no `<use>` inside a span, so `svgIcon.length` is 0 and nothing renders — a permanently blank box, since spans no longer paint a background image.

Checked every current caller of `MenuItem#setIconCls` (`DocumentHolderExt.js`, `Header.js`, etc.) — all of them only ever pass `btn-*` values across a given item's lifetime, so this isn't triggered by any existing code path today. It's a latent trap for the next feature that reuses a `MenuItem` for both an icon and a swatch state, not a live regression.

### [LOW] `apps/common/main/lib/component/MenuItem.js:481-499` — `applyScaling()` wasn't updated for the new markup

Duplicates the same "inject an svg at ratio > 2" logic Button.js has, and was left untouched by the PR (confirmed: zero diff lines touch this function). For a genuine swatch item (non-`btn-*` `iconCls`) at `ratio > 2`, `firstChild.find('svg.menu-item-icon').length` is still 0 (swatches keep the `<span>`), so it builds `<svg class="menu-item-icon uni-scale ..."><use href="#null"></use></svg>` and inserts it right after the swatch span. This is pre-existing behavior (ran the same way before the PR too), so not a new regression, but worth a follow-up guard (skip when `iconCls` has no `btn-*` match).

## What looks correct

- `Common.UI.menuItemIconMarkup()` mirrors `Common.UI.Button`'s existing `btn-*` regex / `<use href="#...">` approach exactly, including the `uni-scale` class that `bigscaling.less`'s `:not(.uni-scale)` rules rely on to keep the icon visible at every DPI tier.
- Non-`btn-*` classes correctly fall back to the old `<span class="menu-item-icon ...">`; the ColorButton "Automatic" swatch and the border-color pickers use their own hand-written templates that never call this helper at all (verified `mnuBorderColor` passes a custom `template` and is never touched by `setIconCls`), so they're unaffected.
- All six call sites the issue named (`MenuItem.itemTemplate`, `Menu`'s default `itemTemplate`, `Header`'s editing-mode menu, Writer/PDF `PageThumbnails` sliders, `FormsTab`/`FormSettings` zoom items) were migrated, and no other manually-duplicated `menu-item-icon` template was missed — the remaining `menu-item-icon` hits in `SideMenu.js`, `Plugins.js`, `Toolbar.js` (border-color `.css()` calls) and `FormatRulesEditDlg.js` are genuinely non-icon usages (thumbnail ``, or literal color swatches) and were correctly left alone.
- `common.less`'s `#file-menu-panel svg.menu-item-icon { display: none → block }` flip matches pre-existing (previously dormant) `.pixel-ratio__2_5` CSS that already anticipated SVG menu icons, confirming the fix is consistent with existing styling rather than inventing a new mechanism.
- `setChecked()`/`toggle()`'s `firstChild.css('background-image', 'none')` targets the `<a>` element, not `.menu-item-icon`, and no CSS rule sets `background-image` on that `<a>` — so this line was already a no-op before the PR and remains harmless after it; checked-state rendering (via the `.checked`/`:before` checkmark) is untouched by the migration.
- `btn-find-redacted` has no matching `<symbol>` in `icons.svg`; the PR doesn't claim to fix that (called out as pre-existing in the issue), and degrades the same way an unmatched `btn-*` class already would (empty `<svg>` instead of empty `<span>`) — not a regression.

## Test gaps

- No automated coverage exists for `Common.UI.MenuItem` / `Common.UI.menuItemIconMarkup` at all. At minimum: a unit test asserting `btn-*` classes produce `<svg><use href="#...">` and non-`btn-*` classes still produce `<span>`.
- No regression test for `setIconCls()` being called with a `btn-*` value after the item was created with a non-`btn-*` value (or vice versa) — this is the MEDIUM finding above and would catch it before it ships.
- No visual/DOM test across the 6 migrated hand-written templates (Header editing-mode menu, both `PageThumbnails` sliders, both zoom menus) confirming the icon actually renders post-migration.
- No high-DPI test exercising `MenuItem.applyScaling` at `ratio > 2` for a swatch-type menu item to confirm it doesn't leave a broken `href="#null"` `<svg>` next to the swatch.

## Overall assessment

The diff is a minimal, targeted port of `Common.UI.Button`'s existing SVG-symbol mechanism onto `Common.UI.MenuItem`, reusing the same `btn-*` regex, the same `uni-scale` class, and the same shipped `icons.svg` sprite rather than inventing a parallel system. All the call sites the issue named were updated, the swatch/color-picker fallback path is preserved untouched, and the CSS change lines up with dormant high-DPI scaffolding that was already anticipating SVG menu icons. The two things worth addressing in a follow-up (not blocking this fix) are hardening `setIconCls()` for icon-kind changes and cleaning up `applyScaling()`'s now-inconsistent svg-injection path — neither is exercised by any current caller, so they don't constitute a live regression against issue #313.

MonaAghili (commented):
Verified against `apps/common/main/lib/component/MenuItem.js` and its callers (not just the diff in isolation). Two things worth fixing before merge, one of them a live bug.

### 🔴 High: `setIconCls` icon-kind mismatch is reachable today, not just latent

`MenuItem#setIconCls` ([MenuItem.js:305-321](../../blob/HEAD/apps/common/main/lib/component/MenuItem.js#L305-L321)) only swaps CSS classes on whatever element currently has `.menu-item-icon` — it never replaces the element itself. Since this PR makes the element type depend on whether the class is `btn-*` (`<svg><use>`) or not (`<span>`), any caller that alternates between the two on the same `MenuItem` instance leaves the wrong tag in the DOM.

This isn't hypothetical. In `DocumentHolderExt.js`, the paragraph-direction submenu item is rebuilt on every context-menu invocation:

- `apps/pdfeditor/main/app/view/DocumentHolderExt.js:1553-1570`
- `apps/presentationeditor/main/app/view/DocumentHolderExt.js` (same pattern)

```js
if (isInShape || isInChart) {
    // ...
    cls = 'menu__icon btn-text-orient-hor'; // etc.
    me.menuParagraphDirection.setIconCls(cls);
} else {
    me.menuParagraphDirection.setIconCls(''); // same instance, non-btn- value
}
```

Repro: right-click a shape/chart (submenu shows the orientation `<svg>` icon correctly), then right-click plain text and reopen the submenu. The old `<svg><use>` is still in the DOM; `setIconCls('')` just writes `href="#null"` onto it (no `btn-` match), leaving a broken/blank icon element instead of a clean empty `<span>`.

`menuTableDirection` (`documenteditor`/`presentationeditor` `DocumentHolderExt.js`) has the same switch-with-`''`-default shape and is plausibly affected too, depending on whether `get_CellsTextDirection()` can return an unmatched value for the current selection — didn't confirm that one as conclusively as the paragraph-direction case above.

Suggested fix: have `setIconCls` rebuild the icon element via `Common.UI.menuItemIconMarkup()` and `.replaceWith()` it, instead of mutating classes on the existing node.

### 🟡 Low: `applyScaling` duplicates `menuItemIconMarkup` instead of calling it

`MenuItem.applyScaling` ([MenuItem.js:481-499](../../blob/HEAD/apps/common/main/lib/component/MenuItem.js#L481-L499)) re-implements the same `btn-` regex + `<svg class="uni-scale"><use></svg>` construction that `Common.UI.menuItemIconMarkup` (added by this PR a few lines above) already does, rather than calling it. It wasn't touched by this diff, so it's not a new regression, but it's now duplicated logic that will drift — e.g. a future change to the markup shape only updates one of the two copies. It also doesn't guard against non-`btn-*` (swatch) items, so at `ratio > 2` a swatch `MenuItem` gets a stray `<svg><use href="#null"></svg>` appended next to its `<span>` (renders empty today, but is dead weight in the DOM).

Suggested fix: have `applyScaling` call `Common.UI.menuItemIconMarkup(me.iconCls)` instead of rebuilding the string inline, and skip injection entirely when `iconCls` has no `btn-` match.

### What looks correct

- `Common.UI.menuItemIconMarkup()` mirrors `Common.UI.Button`'s existing `btn-*` regex / `<use href="#...">` approach, including the `uni-scale` class `bigscaling.less` relies on to keep icons visible at every DPI tier.
- Non-`btn-*` classes correctly fall back to the old `<span class="menu-item-icon ...">`; `ColorButton`'s "Automatic" swatch and the border-color pickers use their own hand-written templates that never call this helper, so they're unaffected.
- All six call sites this PR names (`MenuItem.itemTemplate`, `Menu`'s default `itemTemplate`, `Header`'s editing-mode menu, Writer/PDF `PageThumbnails` sliders, `FormsTab`/`FormSettings` zoom items) were migrated consistently.
- The `common.less` `#file-menu-panel svg.menu-item-icon { display: none → block }` flip is correct and necessary now that the icon lives in the `<svg>` instead of the `<span>`.

test/unit-tests/common/main/lib/component/MenuItem.js
```diff
@@ -0,0 +1,180 @@
...
+ *
+ *  Unit test
+ *
+ *  Covers the icon markup: a sprite name draws an <svg><use> out of the
+ *  shipped icons.svg, anything else stays the <span> the colour pickers
+ *  paint as a swatch.
+ *
+ */
+
+define([
```
juliusknorr: Are we actually running those tests?

juliusknorr (approved):
Seems good to me, just a comment on the test

test/unit-tests/common/main/lib/component/MenuItem.js
chrip: No, nothing runs them: both of the headless-browser test runs we actually have are yours. runAll.js only became a runner that fails a build through your Euro-Office/sdkjs@71440b9f29 (the file:// URL, --no-sandbox, the non-zero exit, the concurrency cap, the documented skip map), and e2e.yml in this repo is yours too, from 6e0d85ed1e.

Here, build.yml builds and your e2e.yml drives the full stack against an overlaid nightly image. test/unit-tests/common/index.html is a page you open by hand. I got it running again in 9324c02 / d0fddf0 in August after it had rotted, but nothing executes it, so this PR adds 25 cases to a suite no runner touches.

The fix is porting what you did in sdkjs. The shape carries over directly: a list of pages, headless Chromium, an exit code. The difference is that this suite is mocha + chai, so node-qunit-puppeteer doesn't fit and the result has to be read off mocha instead. Roughly 40 lines plus a job.

Should I create a new issue for that?

chrip:
@MonaAghili Thanks for checking the callers. I think this review was read against fb7d0fc though, the first commit: both findings were already on the branch, in the two commits that answer your first review.

240ebae428 makes setIconCls rebuild the icon element via Common.UI.menuItemIconMarkup() and replaceWith() it when the kind changes, which is the fix you suggest.

0a22172101 drops the !_.isEmpty(this.iconCls) guard. That guard was the worse half of the DocumentHolderExt bug: setIconCls('') recorded an empty iconCls, so every later call returned before touching the DOM, and the direction icon never came back for the rest of the session. It now guards on being rendered and on an icon element being there, and the lookup is narrowed to span/svg so MenuItemCustom's  plugin icon is never swapped for a symbol.

There's a test for the round trip you describe, "setIconCls clears the icon and takes it back again". It fails against 240ebae428 and passes on 0a22172101.

On the low one: the swatch guard landed in 0a22172101, but you were right that the markup was still assembled twice. d017ea6206 drops applyScaling's inline string and calls the helper. The injected svg now carries the full iconCls instead of only icon-rtl, which is what the template already prints for every other sprite item; sprites take no paint from CSS any more, so the extra classes only make the two paths agree.

Reviewing that commit turned up that your original finding had survived inside the fix for it. wasSprite reads iconEl[0], first in document order. Where applyScaling has injected an svg beside a span, that first element is the span, so a swap to a swatch class looked like no kind change at all: the replace branch was skipped and the leftover svg kept its <use>, pointed at #null. An empty box over the colour, which is exactly what 240ebae428 set out to remove. d017ea6206 collapses to a single element whenever it finds more than one. Nothing in the tree reaches it today, since every hand-written span template carries a non-sprite class (mnu-border-color, color-auto) and the applyScaling guard never fires beside one, so the test builds the pair by hand. It fails against 0a22172101 with "expected 2 to equal 1".

On the test gaps: test/unit-tests/common/main/lib/component/MenuItem.js now has 26 cases covering the markup helper both ways, the setIconCls transitions including the round trip and the pair case, the iconImg survival, and applyScaling at ratio 3 for both a swatch and a sprite. I ran them through headless Chromium to confirm the before and after, since nothing runs them in CI yet (separate thread with Julius). It doesn't cover the six hand-written templates; those I checked by hand in Writer, Spreadsheet and Presentation.

MonaAghili:
Thanks for the trace you're right, I reviewed against `fb7d0fc` and missed that 240ebae/0a22172/d017ea6 already landed on the branch. I checked all three diffs and they resolve both findings as described:

- `setIconCls` rebuilding via `menuItemIconMarkup()`/`replaceWith()` on kind change — confirmed, matches what I suggested.
- The `!_.isEmpty(this.iconCls)` guard removal and narrowed `span/svg` lookup — confirmed, and yes that guard was the nastier bug (icon stuck blank for the session).
- `applyScaling` calling the shared helper, plus the `wasSprite`/multi-element collapse fix — confirmed, and good catch that my Low finding's fix exposed a latent bug in how the kind was detected via `iconEl[0]`.

LGTM! dropping my REQUEST CHANGES.

MonaAghili (approved):

PR #206, opened 2026-08-21
i18n(da): translate Common.Views.Header strings

bargib:
Adds Danish translations for the 20 untranslated keys in Common.Views.Header in documenteditor/main. These were missing entirely — no existing strings are changed.

Danish coverage for this component goes from 67.7% to 68.2% (2,963 → 2,983 of 4,374 keys).

I used the Microsoft Danish localization style guide as reference, and the German translations in the same file for context. Verified that the file parses, key ordering follows en.json, and no placeholders or HTML are involved in these strings.

This is the first of several batches. Danish currently sits at 54.4% across the web editors (rank 40 of 45 locales), and I intend to work through it component by component.

AI disclosure: I used Claude Opus 5 to assist with the tooling and to propose translations. I reviewed and verified each of the 20 strings myself against the English source and the style guide, and I take responsibility for every line.


chrip (approved):
## Summary

Clean first batch. All 20 keys are genuinely new, so nothing existing was touched, and I
confirmed that against the base rather than taking the description's word for it: 20 added, 0
changed, 0 removed. Every translation reads correctly against the English source, with the
German column agreeing throughout. The file parses, the new keys landed in the same order as
`en.json`, and none of these strings carry placeholders or markup, so there was nothing to
break.

Your description is a model of what makes a translation PR reviewable. Stating the coverage
delta, the reference you used, and what you deliberately did not touch turns a wall of unfamiliar
strings into something a non-Danish-speaker can actually check.

## Worth a follow-up, not a change here

Two of the new keys introduce a second Danish word for a term the file already renders
differently elsewhere. Both of your choices look like the better ones, so this is about the old
entries, not yours:

| English | Your new key | Already in the file |
|---------|--------------|---------------------|
| `Download` | `Common.Views.Header.textDownload` → `Download` | `DE.Views.FileMenu.textDownload` and `DE.Views.MailMergeSettings.textDownload` → `Hent` |
| `Print` | `Common.Views.Header.textPrint` → `Udskriv` | `DE.Views.FileMenu.btnPrintCaption` and `DE.Views.Toolbar.tipPrint` → `Print` |

Worth settling once in a later sweep rather than here. The same applies to
`Common.Views.ReviewChanges.txtEditing`, which is lowercase `redigering` where your new
`textEdit` and `tipDocEdit` correctly capitalise it.

## Verdict

**Approve** — 20 correct new translations, nothing existing disturbed, and a description that
made verification quick. Good start to the series.

bargib:
Thanks — agreed on leaving all three for a later sweep.

One correction on the `Download` row, though. You wrote that both of my new choices look like the better ones, and I've since been persuaded that isn't true here: it's the *old* entries that are right. `Hent` should stay, and my new `Common.Views.Header.textDownload` is the one to change in the sweep.

The reference style guide settles it. Its own model translation renders "locate downloaded files" as "finde **hentede** filer", and this file already uses `Hent som` for Download As. `Download` as a Danish loan word is the outlier.

`Print` and `Editing` I read exactly as you do — there the new keys are the better ones and the old entries need fixing.

Flagging it so my sweep doesn't end up going the wrong way at a later date...

chrip (approved):
## Summary

Follow-up to my review of 8 September. The branch has merged `main` since, but the 20 added keys are unchanged and still verify correctly. CI has caught up too: the E2E run was unapproved when I reviewed, and it has since run and passed, so this is green on both checks.

The substance of this follow-up is your comment of 13 September, and you are right.

## You corrected me, and the correction stands

I wrote that both of my flagged rows had your new choice as the better one. For `Download` that was wrong, and your reasoning is sound: `Hent` is the established form and the loanword is the outlier.

I went back to the file rather than just taking your word for it, and the evidence is stronger than either of us put it. Across `documenteditor/main`, English `Download` and its inflections render as `Hent` almost everywhere:

| Key | English | Danish |
|-----|---------|--------|
| `DE.Views.FileMenu.btnDownloadCaption` | Download As | `Hent som` |
| `DE.Views.FileMenuPanels.ViewSaveAs.textDownloadAs` | Download as | `Hent som` |
| `DE.Views.FileMenu.textDownload` | Download | `Hent` |
| `DE.Views.MailMergeSettings.textDownload` | Download | `Hent` |
| `DE.Views.Toolbar.tipDownload` | Download file | `Hent fil` |
| `Common.Views.Header.tipDownload` | Download file | `Hent fil` |

The last row is the one worth noticing: that key is one of the 20 this PR adds. So the PR already commits to `Hent fil` for the tooltip while giving the button itself `Download`. The inconsistency is not only against the old entries, it is inside this PR between a control and its own tooltip. That makes your conclusion firmer, not weaker.

The only stragglers going the other way are `downloadErrorText` (`Download fejlet.`) and `downloadMergeTitle` (`Downloader`), both pre-existing and both outliers on the same reading.

`Print` and `Editing` I still read as you do — there the new keys are right and the old entries need the sweep.

All three rows remain as they were, deliberately: we have agreed to settle them in a later sweep rather than here. The only change is that the `Download` row now has its direction reversed, with `Common.Views.Header.textDownload` as the key to change rather than the old `Hent` entries.

Nothing else moved. Still 20 added, 0 changed, 0 removed.

## Verdict

**Approve**, unchanged. Thank you for coming back to correct the record rather than letting a wrong conclusion sit in an approved review — that is the part that keeps a sweep from going the wrong way months later.

PR #207, opened 2026-08-22
i18n(da): fix Redo mistranslated as Undo

bargib:
I found a bug in the Danish locale where both Undo and Redo were translated as "Fortryd". That is wrong — "Fortryd" is Undo. Redo should be "Gentag".

This is confusing for users, since two different actions end up with the same label on the same toolbar.

I found the same error in 10 strings across 5 components (documenteditor forms and main, pdfeditor, presentationeditor, spreadsheeteditor). Only keys where the English source is exactly Redo are changed — all Undo strings are untouched.

To verify the term I checked Danish Microsoft Word, which labels the two buttons "Fortryd indtastning" (Undo) and "Gentag indtastning" (Redo). German makes the same distinction with Rückgängig and Wiederholen, and the Microsoft Danish style guide lists Ctrl+Y as "Gentag".

AI disclosure: I use Claude Opus 5 to help with tooling and to propose translations. I review every string myself and stand behind each correction.


chrip (approved):
## Summary

This is a real bug and the fix is exactly right. `Fortryd` is Undo, so having it on the Redo
button too meant two adjacent toolbar buttons carried the same label. `Gentag` is the correct
term, and the German pair `Rückgängig` and `Wiederholen` makes the same distinction.

I checked every one of the 10 changes: all 10 target keys have `Redo` as their English source,
all 10 previously read `Fortryd`, and all 10 now read `Gentag`. Nothing else moved. No key was
added or removed, no Undo string was touched, and the file ordering is unchanged.

I also ran the check that matters most for a rename like this, which is whether the fix is
complete and whether it introduced any new split. It did neither: across the files you touched,
`Redo` now maps to `Gentag` and nothing else, and no other English source picked up a second
Danish rendering. That is the clean result.

## One cross-PR note

Pull request #208 adds `Common.Views.MacrosDialog.tipRedo` with the correct `Gentag` while base still has
the old value, so until this PR lands that branch shows `Gentag` and `Fortryd` side by side in
the same file. Merging this one first makes that disappear. Nothing to do here, just worth
knowing the two are complementary.

## Verdict

**Approve** — a genuine mistranslation, fixed completely and precisely, with the reference work
shown. This is the kind of change that should go in quickly.

PR #208, opened 2026-08-22
i18n(da): translate three Common.* namespaces

bargib:
This adds 221 strings across three Common.* namespaces in documenteditor/main:

Common.Utils.ThemeColor — 34 colour names
Common.Views.MacrosDialog — 32 strings
Common.define.smartArt — 155 SmartArt layout names

One commit per namespace, so each can be reviewed separately. No existing strings are changed. Danish coverage for this component goes from 67.7% to 72.7%.

For the SmartArt names I checked Microsoft's published Danish list, since those are established terms Danish users recognise from Office. That page is machine-translated, so I deviated from it in six places where it is plainly wrong — for instance it renders "Balance" as "Saldo", which means an accounting balance. The colour names follow ordinary Danish usage; I did not find a complete authoritative Microsoft reference for those, but did check some of them in Word.

While working here I also found errors in translations that are already in the file. I am not touching them in this PR — I will open a separate correction PR for those.

AI disclosure: I use Claude Opus 5 to help with tooling and to propose translations. I have reviewed every string and every line myself.


chrip (approved):
## Summary

221 new strings and not one existing string disturbed, which I verified against the base:
221 added, 0 changed, 0 removed. One commit per namespace makes this reviewable at the size it
is, and splitting `Common.Utils.ThemeColor`, `Common.Views.MacrosDialog` and
`Common.define.smartArt` that way was the right call.

The SmartArt block is the interesting part, since it takes Danish from 4 translated layout names
to 159. Being explicit that Microsoft's published list is machine-translated, and that you
deviated from it in six places where it is plainly wrong, is exactly the right way to handle a
reference of that quality. `Balance` rendered as `Saldo` is a good illustration of why you
cannot just import that page wholesale.

Structurally everything is sound. No placeholder or markup mismatches anywhere in the 221, no
key ordering drift from `en.json`, and no value accidentally left as its English source except
where that is the correct Danish too.

## Two small things

**`Kopiér` versus `Kopier`.** The new `Common.Views.MacrosDialog.textCopy` uses the accented
form, while the three existing `Copy` keys in the same file read `Kopier` without it. You
address this exact question in #225, where you decide to keep the unaccented form for
consistency with existing occurrences and fix it systematically later. Worth making this key
follow that decision so the two PRs do not pull in opposite directions.

**Missing trailer in the description.** The commits all carry `Assisted-by: Claude:claude-opus-5`,
which is what the policy actually requires, and the prose discloses the assistance clearly. The
description is just missing the trailer line you include in every other PR in this series.

Neither is a reason to hold this up. The `Redo` split my checks flagged is not yours: your new
`tipRedo` is correct at `Gentag`, and the old `Fortryd` next to it is what #207 fixes.

## Verdict

**Approve** — a large, careful batch with honest sourcing and no collateral damage. The
`Kopiér` accent is worth a one-character amendment for consistency with your own later decision,
but it does not block.

chrip (approved):
## Summary

Follow-up to my review of 8 September. The branch has merged `main` since, which matters here because #225 landed on 10 September. The 221 added keys are otherwise unchanged and still verify. CI has caught up: E2E has now run and passed, so this is green on both checks.

**`Kopiér` versus `Kopier` — now sharper, and worth acting on.** When I raised this, the conflict was with a sibling PR. #225 has since merged, so the decision it carried is now in `main`:

| Key | Danish |
|-----|--------|
| `DE.Views.BookmarksDialog.textCopy` (on `main`) | `Kopier` |
| `DE.Views.DocumentHolder.textCopy` (on `main`) | `Kopier` |
| `DE.Views.Toolbar.tipCopy` (on `main`) | `Kopier` |
| `Common.Views.MacrosDialog.textCopy` (this PR) | `Kopiér` |

So this is no longer two open branches disagreeing; it is one new key disagreeing with the settled state of `main`, in the same file, for the same English source. Your own #225 description argued for the unaccented form for exactly this reason, and said the accent question should be fixed systematically rather than piecemeal. Changing this one character would make the PR consistent with the decision you already made and shipped.

Still a one-character amendment and still not a blocker.

**Missing trailer in the description.** Unchanged — the commits all carry `Assisted-by: Claude:claude-opus-5`, which is what the policy requires, and the prose discloses clearly. Only the description line you include in every other PR is absent.

The `Redo` note has partly resolved itself in the sense that it was never yours: your new `tipRedo` is correct, and #207 fixes the old `Fortryd` beside it. #207 is still open.

## Verdict

**Approve**, unchanged. 221 careful additions with honest sourcing on the SmartArt names. The `Kopiér` accent is the one thing I would now actually change before merge, since `main` has since settled the question the other way.

bargib:
On Kopiér vs. Kopier. This is something the style guide is very clear about. If the word in danish is a verb, that can be read as a noun, then we keep the accent, because it makes the word read clearly as a verb. Therefore Kopiér (copy as the action "to copy something") is the right spelling. Kopier could be also in Danish be read as "copies", eg. "I have 5 copies of the text". 

Hope this makes sense... As I work my way through the translations, I get more and more sure about the terms and when to use them. This one I am quite confident in, and has been checked what is common in Danish programs.

PR #210, opened 2026-08-22
i18n(da): translate Common.Controllers.Shortcuts descriptions

bargib:
This translates the 147 keyboard shortcut descriptions in Common.Controllers.Shortcuts for documenteditor/main.

I have not translated the txtLabel* keys in the same namespace. They contain CamelCase identifiers — AddNewRow, BigMoveObjectDown, CenterPara — rather than user-facing text, and most locales have left them alone and they look like internal identifiers that ended up in the locale file.

AI disclosure: I use Claude Opus 5 to streamline the process, but I have reviewed every translation myself.


chrip (approved):
## Summary

The cleanest PR in this series. 147 new keys, 0 changed, 0 removed, and my checks came back
with nothing at all to flag — no placeholder or markup mismatch, no key ordering drift, no
value left sitting as its English source, and not one English string in the file that now
carries two different Danish renderings. For a batch of 147 that is a good result.

Leaving the `txtLabel*` keys alone was the right judgment. They hold CamelCase identifiers like
`AddNewRow` and `CenterPara` rather than user-facing text, so they look like internal names that
leaked into the locale file. Translating them would have produced strings no user ever sees while
making the file harder to reason about. Most locales reaching the same conclusion is good
supporting evidence.

Shortcut descriptions are a good target for this kind of work, too. They are self-contained, they
carry no interpolation, and they are exactly the strings a user hunting for a keyboard shortcut
reads most carefully.

## Verdict

**Approve** — 147 correct additions, nothing existing touched, no issues found on any check I
ran. Merge it.

PR #211, opened 2026-08-22
i18n(da): translate Draw tab across all editors

bargib:
The Draw tab was partially translated in three editors and not at all in presentationeditor. This completes it consistently across all four: documenteditor, pdfeditor, presentationeditor and spreadsheeteditor (27 strings).

The tool names are verified against Danish Microsoft Word, which uses "Pen", "Viskelæder" and "Overstregningstusch" for Pen, Eraser and Highlighter. "Vælg" and "Størrelse" were already present in three of the four editors, so those are unchanged — this brings presentationeditor in line.

Note that Common.Views.Draw.* is duplicated per component rather than shared, so the same eight keys appear in four separate locale files.

AI disclosure: I use Claude Opus 5 to help with tooling and to propose translations. I verified the terminology in Danish Word and reviewed every string myself. 


bargib:
I have now tested this on a local Euro-Office Document Server (ghcr.io/euro-office/documentserver:latest, EXAMPLE_ENABLED, lang=da) and can confirm the Draw tab and its tools render correctly in Danish.

While testing I noticed something worth flagging, though I do not think it is caused by this PR. In presentationeditor, textTabDesign, textTabAnimation and textTabSlideMaster are missing from da.json. On the unmodified image those tabs fall back to English and read "Design" and "Animation". With this PR applied they render as empty buttons instead — the tab is there but has no label.

So the fallback appears to behave inconsistently depending on which neighbouring tabs are translated. The result is that a partially translated tab row can be worse than an untranslated one.

I would like to follow up with a small PR adding those three missing tab names, which should resolve it. Let me know if you would rather have them included here instead.

bargib:
Added one more fix: the tab name "Draw" was translated as "Tegn" (imperative verb, e.g. "draw here"), but Danish tab names are consistently nouns (Hjem, Indsæt, Layout). Corrected to "Tegning" (noun), matching the pattern used in Danish Word and the German translation "Zeichnen" (a substantivized infinitive — the functional equivalent). Note: this does not apply to the two other "Draw" strings in the codebase (PdfSignDialog.txtDraw, DocumentPreview.txtDraw) — those are button/command contexts where "Tegn" as an imperative is correct and were left unchanged.

chrip (approved):
## Summary

27 new keys across the four editors, nothing existing changed, nothing removed. All correct
against the English source, with the German column agreeing. No placeholder or markup issues,
no key ordering drift.

The second commit is the part I want to single out. Translating the tab name as the noun
`Tegning` rather than the imperative `Tegn`, because Danish tab names are consistently nouns,
is the kind of distinction that separates a translation from a word swap. Better still, you
worked out where the imperative *is* correct and left those alone. My checks flagged that
`Draw` now maps to two Danish values per file, and on inspection that split is exactly the one
you describe and exactly right:

| Key | Context | Danish |
|-----|---------|--------|
| `DE.Views.Toolbar.textTabDraw`, `PE.Views.Toolbar.textTabDraw` | ribbon tab name | `Tegning` |
| `Common.Views.PdfSignDialog.txtDraw`, `PE.Views.DocumentPreview.txtDraw` | button command | `Tegn` |

So the divergence is deliberate and context-correct, not an oversight. Verifying the tool names
in Danish Word and then smoke-testing the result in a running Document Server is more diligence
than this size of change strictly needs.

## On the empty tab buttons you flagged

Your observation that a partially translated tab row renders worse than an untranslated one is
a good catch and worth keeping visible beyond this PR. #224 adds the three missing
presentationeditor tab names, so the specific symptom is addressed there. The underlying
behaviour, where the English fallback stops applying once neighbouring tabs are translated,
still looks like a bug in its own right rather than a translation gap. Worth a separate issue
if it reproduces on an unmodified image.

Adding those names in a follow-up rather than expanding this PR was the right instinct.

## Verdict

**Approve** — complete, consistent, verified in a running editor, and the `Tegning` versus
`Tegn` reasoning is sound. Merge it.

PR #212, opened 2026-08-22
i18n(da): fix Formula and Tab mistranslations

bargib:
This fixes 36 mistranslations across five components. No new strings are added. Every change is confirmed against the German translation.

Formula translated as "Formular" (9 occurrences). In Danish, "formular" means a form — the kind you fill in. The word for a spreadsheet formula is "formel". German uses Formel throughout. The worst case is the ribbon tab in spreadsheeteditor, which currently reads "Formular", so a user looking for formulas finds a tab that sounds like it is about paperwork. One of these also has a typo: "Uvententet" for "Uventet".

Tab translated as "Fane" (12 occurrences). In the paragraph settings dialog, Tab means a tab stop, not a ribbon tab. Danish uses "tabulator" for the former and "fane" for the latter — they are different words. German makes the same distinction and uses Tabulator in all of these. "Fane position" is doubly wrong: wrong word plus a missing compound.

Center translated as "Centrum" (3 occurrences) in the tab stop dialog, next to "Venstre" and "Højre". "Centrum" means the centre of a town. The alignment term is "Centreret".

Insert translated as "indsæt" (12 occurrences) with a lowercase initial. All other UI labels are capitalised.

AI disclosure: I use Claude Opus 5 to help with tooling and to propose corrections. I reviewed all 36 changes myself against the English source and the German translation.


chrip (approved):
## Summary

The strongest PR in this batch on the merits. A ribbon tab reading `Formular` when it means
Formula is a genuine usability bug, not a nitpick, and the four corrections are all real
distinctions rather than preferences. I verified all 36 changes against the English source and
the German translation, and every one holds up. No key added, none removed, no placeholder or
markup damage, no ordering drift.

The `Tab` work deserves specific praise, because it is the change most likely to be reverted by
someone who does not look closely. German confirms your split precisely:

| Key | English | German | Your Danish |
|-----|---------|--------|-------------|
| `Common.Views.OpenDialog.txtTab` | `Tab` | `Tabulator` | `Tabulator` |
| `SSE.Controllers.Main.txtTab` | `Tab` | `Registerkarte` | `Fane`, unchanged |

Same English word, two different concepts, and you changed exactly the one that means a tab
character while leaving the one that means a ribbon tab alone. That is the right call and it is
worth stating in the PR so nobody 'fixes' it later.

## Still there: ten more `Centrum`

You fixed `Centrum` in the tab stop dialogs, where it sat next to `Venstre` and `Højre` and was
clearly wrong. The same word remains for the same English source in ten other keys, all of them
alignment or position contexts where `Centreret` would apply for the same reason:

| File | Keys still reading `Centrum` |
|------|------------------------------|
| `apps/documenteditor/main/locale/da.json` | `DE.Views.DocumentHolder.centerText`, `DropcapSettingsAdvanced.textCenter`, `ImageSettingsAdvanced.textCenter`, `ListSettingsDialog.textCenter`, `TableSettingsAdvanced.textCenter`, `TableSettingsAdvanced.textCenterTooltip` |
| `apps/presentationeditor/main/locale/da.json` | `PE.Views.DocumentHolder.centerText` |
| `apps/spreadsheeteditor/main/locale/da.json` | `SSE.Controllers.DocumentHolder.centerText`, `SSE.Views.ChartSettingsDlg.textCenter`, `SSE.Views.HeaderFooterDialog.textCenter` |

All three files are already in this PR, so these are in-place edits rather than new territory.
Equally fine as a follow-up if you would rather keep the scope as described — it does not need
to hold up the merge.

## Verdict

**Approve** — four real mistranslations fixed correctly, and the `Tabulator` versus `Fane`
distinction is precisely reasoned. The remaining `Centrum` occurrences are a nice-to-have.

PR #214, opened 2026-08-25
fix(license): stop enforcing licence restrictions in the editors

### LINKED ISSUE #93: [Mobile] EuroOffice crashes on opening documents - TypeError: undefined is not an object (evaluating 't.warnNoLicense.replace')⁠
Hello!
I recently migrated from a bare-metal Nextcloud installation to Nextcloud AIO. During the setup, I chose EuroOffice as my document editor.
On Desktop, everything works perfectly fine. However, whenever I try to open any document on a mobile device (both via the mobile web browser and the official Nextcloud mobile app), the editor fails to load and results in an empty/undefined screen.
The exact same thing happens with text documents, spreadsheets, and presentations.
**Error Logs:**
Upon inspecting the mobile developer console, I found a fatal JavaScript error related to license checking that seems to crash the entire mobile app interface:
```javascript
[Error] TypeError: undefined is not an object (evaluating 't.warnNoLicense.replace')
applyLicense (app.js:9:1741156)
r (app.js:9:1734340)

```
**Environment details:**
 * **Nextcloud Version:** Nextcloud AIO (latest)
 * **Reverse Proxy:** Nginx Proxy Manager
 * **Office App:** EuroOffice
 * **Client:** Mobile web browsers (Safari/Chrome) and Nextcloud Mobile App (tested on iOS)

Below you can find screenshots of the developer console with the full stack trace. Let me know if you need any additional logs from my container!

Image

### LINKED ISSUE #161: [Mobile] EuroOffice crashes on opening documents - TypeError: undefined is not an object (evaluating 't.warnNoLicense.replace')
I created issue on the nextcoud-eurooffice but no one replay, so maybe here someone would help me with this problem.
Here is like to [issue](https://github.com/Euro-Office/eurooffice-nextcloud/issues/93).

bnjamin:
Resolves #161. Resolves Euro-Office/eurooffice-nextcloud#93. Supersedes #185.

## The problem

On mobile, opening any document crashes in locales that lack the licence-warning strings — Ukrainian and Polish in particular. `applyLicense()` began with:

```js
const warnNoLicense = _t.warnNoLicense.replace(/%1/g, __COMPANY_NAME__);
```

Those three `.replace()` calls run *before* any branch, so a locale missing the string throws on every document open regardless of licence state. Presentations simply failed to open for Nextcloud users in those languages.

## The approach

#185 proposed defaulting the missing strings to their key names. That fixes the symptom but keeps a feature Euro-Office should not have: the block those strings belong to is commercial licence enforcement. It renders *"You've reached the limit for simultaneous connections to %1 editors"* with **Buy Now** → `PUBLISHER_URL` and **Contact Us** → `SALES_EMAIL`, and on the way disconnects co-authoring and deactivates the edit controls. The trigger is `ConnectionsOS` / `UsersCountOS` — the open-source-build seat cap.

Euro-Office is an AGPL fork and should never enforce anything via a licence, so this deletes the enforcement rather than papering over it. The crash and the upsell go together.

Removed across the four mobile editors, the five desktop editors, forms and the four embed viewers:

- the `_state.licenseType` branch of `applyLicense()`, including the buy-now dialog and the `asc_coAuthoringDisconnect()` / `toolbar:deactivateeditcontrols` it triggered
- `onLicenseChanged()` and its `asc_onLicenseChanged` registration
- the `Expired` / `Error` / `ExpiredTrial` / `NotBefore` / `ExpiredLimited` gate in `onEditorPermissions`, which refused to open the document at all
- the "Paid feature" custom-loader dialog — unreachable since f294a2baa6 unlocked `canBrandingExt`: it requires `!canBrandingExt` while also reading `customization.loaderName`, which cannot both hold
- `canLicense`, which was ANDed into `isEdit`, `canComments`, `canReview`, `canChat`, `canUseHistory`, `canFillForms`, `canPDFEdit` and the rest — permissions now come from the host config alone
- `canBranding` (`asc_getCustomization`), which hard-gated integrator branding, the header logo and the About customer panel
- the licence term of `canLiveView` (`asc_getLiveViewerSupport`) — a view quota, `connectionsView > 0 || usersViewCount > 0`. The capability logic is kept: `canLiveView` still requires view mode, and the co-authoring mode, `tr.live-viewer` settings row and header user count are untouched
- `permissionsLicense`, `trialMode`, the `licenseType` state fields, the `licType` parameter thread through `setPermissionOptions`, and 19 strings left unreferenced

`canBranding` and `canLiveView` were confirmed licence-derived by reading `Euro-Office/sdkjs` `common/apiBase.js:_onEndPermissions` and `Euro-Office/server` `DocService/sources/DocsCoServer.js` + `Common/sources/utils.js:1179`, rather than inferred from their names.

**Kept deliberately:** the open-source-build notice; the anonymous-access restriction, which is *not* licence-derived (`DocsCoServer.js` reads it from server config `services.CoAuthoring.server.isAnonymousSupport`, default `true`); and the LeftMenu Trial/Developer badge.

## Why this is one commit

I first split the code across two commits (enforcement, then branding/live-view). An independent review showed that boundary is not real: the enforcement branch is the *only* consumer of `permissionsLicense` and `trialMode`, the paid-feature dialog is the `else if` immediately following it in the same chain, and the branding/live-view lines sit inside the same `setPermissionOptions` bodies. Splitting by diff hunk produced commits that were individually inconsistent — one left the PE/SSE/Visio embed licence gate in place while deleting the strings it referenced. It is now one atomic commit; the mechanical locale sweep and the test-harness alignment remain separate.

## Verification

`node scripts/build-pipeline.js` equivalents run locally: all six desktop webpack configs, all four mobile builds and `deploy-embed` compile with zero errors, and `verify-bundles` / `verify-replacements` / `verify-browser-target` all pass. The existing mocha unit harness is green (12 passes, 0 failures). Grepping the built bundles confirms `canLicense`, `canBranding`, `asc_getLiveViewerSupport`, `warnNoLicense`, `textPaidFeature` and `c_oLicenseResult` are absent, while `warnLicenseAnonymous` and the Trial/Developer `c_oLicenseMode` badge survive in exactly the expected files.

This PR was also reviewed adversarially by an independent agent before submission; it caught a real regression (see below) which is fixed here.

## Known gap: no new tests

CONTRIBUTING asks for tests with every bug fix and this PR has none — flagging rather than hiding it. The crash was a dereference of a missing locale string in code this PR deletes, so there is nothing left to regression-test directly. The options I see are a guard test asserting the removed keys stay absent from the locale files, or a unit test on `setPermissionOptions` proving permissions derive from config alone. Neither fits the browser-mocha harness especially well. Happy to add either if reviewers prefer.

Note that the regression found in review — `canEdit` silently losing its `isSupportEditFeature` term — was invisible to both the build and a syntax check, because the result was still valid JavaScript. That is an argument for the `setPermissionOptions` unit test if reviewers want one.

## Stack

1. **this PR** — remove licence enforcement and gating
2. bnjamin/web-apps#4 — drop the 5723 orphaned translation keys
3. bnjamin/web-apps#3 — align the test integrator pages

Both are queued as drafts on my fork and will be retargeted here as this lands.

### DIFF
--- a/apps/common/main/lib/controller/Chat.js
+++ b/apps/common/main/lib/controller/Chat.js
@@ -114,7 +114,7 @@ define([
         },
 
         onUsersChanged: function(users, currentUserId){
-            if (!this.mode.canLicense || !this.mode.canCoAuthoring) {
+            if (!this.mode.canCoAuthoring) {
                 var len = 0;
                 for (name in users) {
                     if (undefined !== name) len++;
--- a/apps/common/mobile/lib/controller/collaboration/Collaboration.jsx
+++ b/apps/common/mobile/lib/controller/collaboration/Collaboration.jsx
@@ -26,7 +26,7 @@ class CollaborationController extends Component {
         const appOptions = this.props.storeAppOptions;
         /** coauthoring begin **/
         let isFastCoauth;
-        if (appOptions.isEdit && appOptions.canLicense && !appOptions.isOffline && appOptions.canCoAuthoring) {
+        if (appOptions.isEdit && !appOptions.isOffline && appOptions.canCoAuthoring) {
             // Force ON fast co-authoring mode
             isFastCoauth = true;
             api.asc_SetFastCollaborative(isFastCoauth);
--- a/apps/common/mobile/lib/view/About.jsx
+++ b/apps/common/mobile/lib/view/About.jsx
@@ -8,8 +8,7 @@ const PageAbout = props => {
     const { t } = useTranslation();
     const _t = t("About", { returnObjects: true });
     const store = props.storeAppOptions;
-    const isCanBranding = store.canBranding;
-    const licInfo = isCanBranding ? store.customization : null;
+    const licInfo = store.customization;
     const customer = licInfo ? licInfo.customer : null;
     const {
         name: nameCustomer = null,
--- a/apps/documenteditor/embed/js/ApplicationController.js
+++ b/apps/documenteditor/embed/js/ApplicationController.js
@@ -682,25 +682,9 @@ DE.ApplicationController = new(function(){
     }
 
     function onEditorPermissions(params) {
-        var licType = params.asc_getLicenseType();
-        if (Asc.c_oLicenseResult.Expired === licType || Asc.c_oLicenseResult.Error === licType || Asc.c_oLicenseResult.ExpiredTrial === licType ||
-            Asc.c_oLicenseResult.NotBefore === licType || Asc.c_oLicenseResult.ExpiredLimited === licType) {
-                common.controller.modals.showWarning({
-                    title: Asc.c_oLicenseResult.NotBefore === licType ? me.titleLicenseNotActive : me.titleLicenseExp,
-                    message: Asc.c_oLicenseResult.NotBefore === licType ? me.warnLicenseBefore : me.warnLicenseExp,
-                    buttons: []
-                });
-        
-                $('#dlg-warning').css('z-index', 20002);
-                $('#dlg-warning button.close, #dlg-warning .modal-footer').remove();
-            return;
-        }
-
-        appOptions.canLicense     = (licType === Asc.c_oLicenseResult.Success || licType === Asc.c_oLicenseResult.SuccessLimit);
         appOptions.canFillForms   = false; // use forms editor for filling forms
-        appOptions.canSubmitForms = appOptions.canLicense && (typeof (config.customization) == 'object') && !!config.customization.submitForm;
-        appOptions.canBranding  = params.asc_getCustomization();
-        appOptions.canBranding && setBranding(config.customization);
+        appOptions.canSubmitForms = (typeof (config.customization) == 'object') && !!config.customization.submitForm;
+        setBranding(config.customization);
 
         var type = /^(?:(docxf|oform))$/.exec(docConfig.fileType);
         appOptions.isOForm = !!(type && typeof type[1] === 'string'); // oform and docxf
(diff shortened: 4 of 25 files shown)


bnjamin:
### Pre-submission review findings, and what changed

I had this branch reviewed adversarially by an independent agent before asking for human review. It found two blockers, both real, both now fixed. Recording them here so reviewers can see what was checked.

**P1 — `canEdit` lost its `isSupportEditFeature` term** (`apps/documenteditor/mobile/src/store/appOptions.js`, `apps/presentationeditor/mobile/src/store/appOptions.js`)

The clause I removed, `(!this.isReviewOnly || this.canLicense) &&`, sat in the *middle* of a multi-line expression. My edit terminated the statement one line early:

```js
this.canEdit = (permissions.edit !== false || permissions.review === true) &&
    (this.config.canRequestEditRights || this.config.mode !== 'view');   // ← statement ended here
    isSupportEditFeature;                                                 // ← dangling no-op
```

`canEdit` therefore no longer required `isSupportEditFeature`, which would have allowed edit mode on builds where edit features are unsupported. Still valid JavaScript, so neither the webpack build nor `node --check` caught it. Fixed by restoring the `&&`. The desktop equivalent in `documenteditor/main` was unaffected — there the removed clause genuinely was the last term.

**P1 — commits were individually inconsistent.** I had split the code change across two commits by diff hunk. That put the PE/SSE/Visio embed licence-gate removal in the second commit while the deletion of the strings it references stayed in the first, so this PR alone would have rendered a warning dialog from `undefined` values. The code change is atomic and is now one commit; details in the PR description.

**Non-blocking, also fixed:** dead `trialMode` assignment in `documenteditor/forms`, and an unused `isForm` local in `documenteditor/mobile`'s `applyLicense()`.

**Non-blocking, not fixed:** `applyLicense()` is now a misleading name — it only handles the open-source-build notice and the anonymous-access restriction. I left it alone to keep this diff to behaviour rather than renames; happy to rename it in a follow-up if reviewers would like.

The reviewer confirmed the retained anonymous / open-source / Trial-Developer behaviour and the remaining permission formulas were consistent, and agreed no additional test blocker is needed beyond the disclosed absence.

bnjamin:
Renamed `applyLicense` → `applyRestrictions` in a follow-up commit (c366600824), closing the last non-blocking item from the pre-submission review.

Nothing in that method concerns a licence any more — what survives is the open-source-build notice and the anonymous-access restriction, and the latter is driven by server config, not a licence. Pure rename: 25 references, 10 definitions and 15 call sites across the four mobile editors, the five desktop editors and forms. No behaviour change.

Kept as its own commit so the behaviour change stays reviewable separately from the rename. All ten bundles rebuild clean, the three verify gates pass, and `applyLicense` no longer appears in any built bundle.

PR #209, opened 2026-08-22
i18n(da): fix spelling errors in existing translations

bargib:
This fixes 25 spelling errors in existing Danish translations across six components. No new strings are added.

Most are verifiable without knowing Danish:

Certificate reads "Cerfitikant" — transposed letters; should be "Certifikat"
Change password reads "Skrift kodeord" — "skrift" means font, so the label currently says roughly "font password". Should be "Skift kodeord"
Preview reads "Forhåndvisning" in 11 places — missing an "s"
Accent process reads "(intet)", Danish for "none" — a placeholder that was never filled in
Preview results reads "Forhåndsivs resultater " — transposed letters plus a trailing space

One of these looks like a copy from the German file rather than a typo: Comments reads "Kommentare", which is correct German but not Danish (should be "Kommentarer"). It is already correct in presentationeditor, so this also removes an inconsistency between components.

Each change matches on both the English source string and the exact current Danish value, so nothing similar-looking is altered by accident.

AI disclosure: I use Claude Opus 5 to help with tooling and to propose corrections. I have reviewed all 25 changes myself against the English source and the German translation.

Assisted-by: Claude:claude-opus-5

PR #219, opened 2026-09-02
i18n(da): translate Animation tab and effect names

bargib:
This adds Danish translations for the Animation tab in presentationeditor: the 33 PE.Views.Animation.* / PE.Views.AnimationDialog.* control strings (Duration, Delay, Start, Repeat, Trigger, etc.) plus 199 Common.define.effectData.* strings naming animation effects, motion paths, shapes and directions.

Of these, 230 keys were entirely missing from the Danish locale. One existing entry (textPathLine) is corrected from the outdated spelling "Linie" to the current "Linje", per Retskrivningsordbogen.

I have read through every one of the 232 entries myself and checked them against publicly available references: the common Danish Localization Style Guide where it applies (decimal comma, terminology for colors and directions), and the German translation (99.7% coverage) as a second reference point for PowerPoint-specific animation and motion-path names not covered by the style guide.

A handful of entries showed a genuine mismatch between the English source text and the German translation: textPentagon, textSpring, textDiagonalUpRight, and textToFromScreenBottom. In each case the German string describes something different from what the English key/text suggests (e.g. textPentagon → German "direction arrow", not a pentagon), which looks like shifted shape indices somewhere upstream rather than a translation choice on either side. I translated these to match the English source, verified them visually in a running instance, but would still welcome a second pair of eyes on these four specifically.

Verified via smoketest against a local Euro-Office Document Server (ghcr.io/euro-office/documentserver:latest, lang=da): all 231 changed keys render correctly, including the Animation pane controls and a sample of the effect names in the "More effects" dialog.

AI disclosure: I used Claude Opus 5 to help build the translation batch (cross-referencing English/German/style guide) and the script that applies it to the locale file. I reviewed every entry myself, made corrections where I judged the automated suggestion needed adjusting, and take responsibility for the final result.


chrip (approved):
## Summary

231 keys, of which 230 are new and one is the `Linie` to `Linje` correction on `textPathLine`.
I confirmed that split against the base, and the structural checks are all clean: nothing
removed, no placeholder or markup mismatch anywhere in the 231, no key ordering drift from
`en.json`, and no value left sitting as its English source.

Flagging the four entries where the English source and the German translation genuinely disagree,
instead of quietly picking one, is the right way to handle that. It is also what made this
reviewable, so I spent my time there.

I brought Swedish in as an independent third reference, since it was translated separately from
German. Three of your four choices are confirmed, and Swedish agrees with you against German in
every one of them:

| Key | English | German | Swedish | Your Danish | Verdict |
|-----|---------|--------|---------|-------------|---------|
| `textPentagon` | `Pentagon` | `Richtungspfeil`, a direction arrow | `Femhörning`, a pentagon | `Femkant` | ✓ German is the outlier |
| `textDiagonalUpRight` | `Diagonal up right` | `Diagonal, in linke untere Ecke`, the opposite corner | `Diagonalt upp höger` | `Diagonalt op til højre` | ✓ German is the outlier |
| `textToFromScreenBottom` | `Out to screen bottom` | `Verkleinern nach unten`, shrink downward | `Ut till skärmens nederkant` | `Ud til bunden af skærmen` | ✓ German is the outlier |
| `textSpring` | `Spring` | `Spirale` | `Fjädra`, to spring or bounce | `Spiral` | ⚠️ worth revisiting |

`textSpring` is the exception, and it goes the other way from what you describe. You wrote that
you translated these four to match the English source, but `Spiral` follows the German reading
of the word as a coil, while Swedish reads the English as the verb and renders it as a bounce.
Since Swedish and German disagree here and you have seen the effect run, your visual check is
better evidence than either of them. It is only worth a sentence in the PR saying the animation
really does look like a coil, so the choice does not read as an oversight later.

## One thing to keep an eye on

Several direction names now read one way in the Animation pane and another way in the
Transitions pane for the same English source, because the transitions were translated earlier
and separately. `From left` is `Fra venstre` in your new strings and `Venstre` in the existing
ones, and the same pattern holds for the other seven directions, plus `Fade`, `Wipe` and
`Collapse`. Two separate galleries, so this may well be fine as it stands, and it is certainly
not a reason to hold this PR. Worth unifying whenever you next pass through
`PE.Views.Transitions`.

Two others resolve on their own. `Preview` shows as both spellings only because base still has
the misspelling that #209 and #224 fix, and your lowercase `s` for seconds is correct against
the existing capitalised `S`.

## Verdict

**Approve** — 231 careful entries, honest flagging of the four ambiguous ones, and verified in
a running editor. Three of the four check out against an independent third language; `textSpring`
just needs a note rather than a change.

PR #220, opened 2026-09-02
i18n(da): translate slide, image, shape and table settings labels

bargib:
This adds Danish translations for eight labels in the presentationeditor settings panels: the slide background controls (Show background graphics, Reset background, Apply to all slides), the image opacity slider, and the "General" tab name in the advanced settings dialogs for images, shapes, tables and charts.

A note on "Opacity": the key is named strTransparency while the source text reads "Opacity". The three existing Danish entries with the same source text (ShapeSettings, SlideSettings, TextArtSettings) already use "Gennemsigtighed" (transparency), and I've matched that for ImageSettings to keep the UI consistent. "Uigennemsigtighed" would be the more literal translation, but it's cumbersome in Danish and the slider's effect is immediately visible, so I've prioritised consistency and readability. Happy to revisit if the project prefers the literal term across all four.

Not included: Common.Views.FormatSettingsDialog is entirely untranslated in Danish (31 keys, in this and two other editors). I've left it alone rather than translating a single string from it, which would leave a half-translated dialog. I intend to follow up with that as a separate contribution.

Verified via smoketest against a local Euro-Office Document Server (lang=da).

AI disclosure: Claude Opus 5 assisted with locating the missing keys and the apply script; I reviewed each string and verified the result in a running instance.


chrip (approved):
## Summary

Eight new keys, nothing existing changed, nothing removed, and every check clean. All eight read
correctly against the English source with German agreeing, no placeholders involved, and the keys
landed in `en.json` order.

The two judgment calls in the description are both the right ones, and more importantly you
showed your work on them. Matching `Gennemsigtighed` for `strTransparency` because the three
existing keys with the same source text already use it puts UI consistency ahead of a more
literal rendering, which is correct when the slider's effect is visible anyway. And leaving
`Common.Views.FormatSettingsDialog` alone rather than translating one string out of 31 is exactly
the right instinct. A half-translated dialog reads worse than an untranslated one, which is the
same effect you documented on the empty tab buttons in #211.

## One cross-PR note

The second commit gives `strApplyAllSlides` the loanword form `Anvend på alle slides`, while the
existing `PE.Views.Transitions.txtApplyToAll` still reads `Anvend på alle dias` for the same
English source. That is not a gap in this PR: #224 migrates that key to `slides` as part of its
63-string `dias` to `slide` sweep, so the two converge once both land. Merge order does not
matter, since neither depends on the other. Nothing to do here.

## Verdict

**Approve** — small, correct, well-reasoned, and smoke-tested. The scoping decision to skip the
31-key dialog is the part I would most want other contributors to copy.

PR #224, opened 2026-09-06
i18n(da): translate presentation tabs and modernise slide and linje spelling

bargib:
Four related pieces of work in presentationeditor, plus two spelling
normalisations that reach the other editors. 304 strings across 7 files: 72 added and 232 changed.

**Primary changes**

- **Missing tab names (3).** Closes the follow-up I promised in #211 —
  textTabDesign, textTabAnimation and textTabSlideMaster were absent, so those
  tabs rendered as empty buttons rather than falling back to English.
- **View and Slide Master tabs (69).** PE.Views.ViewTab and
  PE.Views.SlideMasterTab were entirely untranslated.
- **"dias" → "slide" (63).** Common presentation tools in danish keep the loanword: the ribbon
  reads "Ny slide" and the group is "Slides". Common gender, English plural
  ("alle slides"). 
- **"linie" → "linje" (155).** Continues the textPathLine correction from
  #219, applied consistently across the four editors. "linje" is the current
  form per Retskrivningsordbogen.

The remaining edits are consequential changes and polish within the same
strings: compound spacing (Slidenummer, Linjeafstand), a typo ("Insæt" →
"Indsæt"), initial capitalisation, "horisontal" where the same file used
"vandret" two menu items away, and one mistranslation — txtSlideAlign
("Align to Slide") read "Tilpas til" rather than "Juster til".

**Worth a second opinion**
I'm fairly sure about these, but chime in, if you want:
- `strLineWeight` → "Linjetykkelse"; "Stregtykkelse" is also defensible.
- `textTabStyle` → "Fanestil"; "Faneformat" would also work.
- `txtOverbar` / `txtUnderbar` → "Streg over/under tekst" and
  `txtShowBottomLimit` → "Vis nedre grænse". Both previously used "linje",
  which is wrong: the first is equation notation, the second a limit.
- `textTabAnimation` kept as "Animation" (singular) to match our source.
  Danish common presentationsprogram has "Animationer", but its source is "Animations".

**Conflicts:** none. I checked every key against my nine other open danish translation PRs; the
only overlap is textPathLine, where #219 makes the same change.

**Verification:** JSON parses, key order follows en.json, placeholders and
HTML untouched. Smoke-tested against ghcr.io/euro-office/documentserver:latest
(lang=da) — all 304 keys applied and render correctly.

AI disclosure: I used Claude Opus 5 for the translation batch and tooling. I
reviewed all 304 strings myself against English and German side by side and
corrected the suggestions where my reading differed — including
"Slideundertitel", where the suggestion had "undertekst". I take full
responsibility for the result.


bargib:
The final commit corrects three strings that had kept their outdated values
from main (textJustify, insertText, textPreview). #209 fixes the same three,
identically, so the two branches now agree rather than conflicting textually.

bargib:
Added one more commit: `capBtnSelect` ("Select", the selection tool in the
View tab) was translated as "Marker". The reference Danish style guide lists
"markér" among the imperatives that always take the accent, because the form
can otherwise be misread as a noun — the guide's own shortcut table renders
Edit > Select All as "Markér alt". Corrected to "Markér".

This does not change the string count: the key is one of the 69 View tab
strings this PR adds, so only its value changed.

I also checked the rest of the PR against both of the style guide's rules on
imperatives — the accent rule above, and the advice to avoid imperatives
ending in consonants ("Ændr", "Lagr", "Forenkl"). "Markér" was the only hit
across all 293 strings.

chrip (approved):
## Summary

The largest PR in the series and still easy to follow, because the eight commits each do one
thing. I verified the whole thing structurally: 72 keys added, 231 changed, none removed, no
placeholder or markup mismatch across all 303, no key ordering drift in any of the seven files,
and every file still parses. The `linie` to `linje` sweep and the `dias` to `slide` migration are
both applied consistently rather than half-done, which is the failure mode I was looking for.

Your conflict check was also right where it counts. I compared every touched key against all ten
sibling branches: four keys overlap, and all four carry identical values, so nothing conflicts
textually or semantically no matter what order these merge in. One small attribution slip, since
`DE.Views.DocumentHolder.insertText` overlaps with #212 rather than #209, but the substance of
the claim holds.

| Key | English | German | Your Danish | Verdict |
|-----|---------|--------|-------------|---------|
| `SSE.Views.ChartSettings.strLineWeight` | `Line weight` | — | `Linjetykkelse` | ✓ and it is the coherent pick, since `Stregtykkelse` would sit oddly inside a `linie` to `linje` sweep |
| `PE.Views.ViewTab.textTabStyle` | `Tab style` | `Stil der Registerkarte` | `Fanestil` | ✓ German confirms this is the ribbon tab, so `fane` is the right stem |
| `PE.Views.DocumentHolder.txtOverbar` / `txtUnderbar` | `Bar over text` / `Bar under text` | `Linie über dem Text` | `Streg over tekst` / `Streg under tekst` | ✓ and better than German, which reads the bar as a line where English says bar |
| `PE.Views.Toolbar.textTabAnimation` | `Animation` | `Animation` | `Animation` | ✓ matches source and German exactly |

`txtSlideAlign` is worth mentioning too, even though you did not ask. `Align to Slide` reading
`Tilpas til dias` was a real mistranslation, and `Juster til slide` matches the German
`An Folie ausrichten`.

## Small gaps, none blocking

**`txtShowBottomLimit` only got fixed in one editor.** The description says the string previously
used `linje`, which is wrong for a limit. That correction landed in
`apps/documenteditor/main/locale/da.json`, but presentationeditor still reads `Vis nederste linje`
even though it is this PR's main target. Spreadsheeteditor already said `grænse` and is fine.

**The title does not match the rest of the series.** `I18n da/presentation tab names and more` is
the branch name rather than a commit subject, and "and more" hides four distinct pieces of work.
Every other PR you have open uses `i18n(da): …`, and since a squash merge takes the title as the
commit message, this one would land on `main` as the odd one out. Something like
`i18n(da): add presentation tab names, translate View and Slide Master, normalise linje and slide`
would match.

**The count is slightly under.** The description says 290 strings and your comment says 293. The
actual figure is 303, being 72 added and 231 changed.

**Two freebies while you are in `textCompactView`.** It carries a trailing space, inherited from
base rather than introduced here, and its English source is `Hide Toolbar` while the Danish reads
`Vis kompakt værktøjslinje`, which says the opposite. Neither is your doing and both are one-line
fixes in a string you are already touching.

**`Brugerdefineret` capitalisation.** Your new `PE.Views.ViewTab.textCustom` correctly
capitalises, in line with your own initial-capitalisation commit, but `PE.Controllers.Main.txtSldLtTCust`
and `PE.Views.SlideSizeSettings.txtCustom` still read `brugerdefinerede`.

On `Markér` versus `Vælg`: your accent reasoning is sound and I checked the style guide point you
cite. Note the split is not only about the accent, since `Select` is `Vælg` in five other keys in
the same file. The selection tool and a texture picker are arguably different actions, so this
looks intentional rather than wrong.

## Verdict

**Approve** — 303 keys of consistent, well-sequenced work, smoke-tested, with an accurate conflict
analysis and four sound terminology calls. The title is the one thing I would genuinely change
before merging, since it outlives the PR.

chrip (approved):
## Summary

Follow-up to my review of 8 September. The branch has merged `main` since and the title has been rewritten. The 303 changed keys are otherwise unchanged. CI has caught up, with E2E now run and passed, so this is green on both checks.

**Title — fixed.** It now reads `i18n(da): translate presentation tabs and modernise slide and linje spelling`, which follows the convention the rest of the series uses and names all the work rather than hiding it behind "and more". This was the one point I said I would genuinely want changed before merge, so that is settled.

**`txtShowBottomLimit` — still one editor short.** Unchanged at head:

| File | Key | Danish |
|------|-----|--------|
| `apps/documenteditor/main/locale/da.json` | `DE.Views.DocumentHolder.txtShowBottomLimit` | `Vis nedre grænse` ✓ |
| `apps/presentationeditor/main/locale/da.json` | `PE.Views.DocumentHolder.txtShowBottomLimit` | `Vis nederste linje` |

The description says the string previously used `linje`, which is wrong for a limit. That correction landed in documenteditor but not in presentationeditor, which is this PR's main target. Spreadsheeteditor already reads `grænse` and is fine. Still a one-line gap.

**Key count — still understated.** The description says 290 and your comment says 293. Measured against the merge base the figure is 303: 72 added and 231 changed. Worth correcting since the description now doubles as the merge commit message.

**`textCompactView` — unchanged.** Still `Vis kompakt værktøjslinje ` with the inherited trailing space, and its English source is still `Hide Toolbar` while the Danish says the opposite. Neither is your doing; both are one-line fixes in a string you are already touching.

**`Brugerdefineret` capitalisation — unchanged.** `PE.Controllers.Main.txtSldLtTCust` and `PE.Views.SlideSizeSettings.txtCustom` still read `brugerdefinerede` while your new `PE.Views.ViewTab.textCustom` capitalises.

Worth noting that #227 now touches the same `Custom` divergence from the other side, so whichever of the two lands first, the other is the natural place to finish it.

**Merge safety.** This branch shares four keys in total with #209, #212 and #219, all with identical values, so order does not matter anywhere. Across all fifteen open Danish PRs there is no key that two branches set to different values.

## Verdict

**Approve**, unchanged. The title fix was the item that mattered most and it is done. The rest are small and stay non-blocking, though `txtShowBottomLimit` in presentationeditor is the one I would still fold in, since the description already claims it.

PR #225, opened 2026-09-06
i18n(da): propagate Common.* strings to presentationeditor and align terminology

bargib:
107 already-translated Common.* strings are propagated into
presentationeditor, and the terminology they touch is aligned across the other
editors. 214 strings in 11 files — 111 new translations and 103 corrections.

**Primary changes**
- **Propagation (107).** Common.* strings that already exist in Danish
  elsewhere, brought into presentationeditor. 26 of them deviate from the
  donor because the donor was wrong — including one that would have copied
  the placeholder "(intet)" into a third editor.
- **`Find` → `Søg` (26).** The Danish style guide's shortcut table
  renders Edit > Find (Ctrl+F) as "Søg". German uses "Suchen", Swedish "Sök".
- **`Equation` → `Ligning` (10).** English itself distinguishes the object
  (Insert > Equation) from a named formula (Quadratic Formula); Danish should
  too. German has Gleichung/Formel, Swedish Ekvation. "formel" is kept for the
  named formulas.
- **`Accent Bar` → `markeringslinje` (24).** Left untranslated inside the
  callout shape names. German renders it "Markierungsleiste"; Swedish left the
  whole string in English.

The rest are consequential changes and polish in the same strings: storage
rendered four different ways collapsed to "lager", English title case removed
from 21 Danish strings, "Case sensitive" aligned with the style guide's own
wording, one string that had lost half its sentence, and "Snak" → "Chat" where
the same key is "Chat" in six other places.

Four came out of smoke testing the running editor: "Vis fra nuværende billede"
used "billede" (image) where the source says Slide, "Insert Spreadsheet" was
untranslated, and the animation trigger read "Start på klik" where #219 renders
the sibling triggers "Ved klik".

**Might be worth a second opinion**
- `markeringslinje` for "accent bar" — chosen to match "markering" used
  elsewhere for accent, but German's "Markierungsleiste" is the only real
  precedent.
- `Billede fra lager` for "Image from storage" — "lager" is the established
  stem (11 existing "Fra lager"), but "opbevaring" and "filarkiv" also occur.
- `Kopier` is left without the accent and `Indsæt` is kept for Paste, both for
  consistency with the ~11 existing occurrences of each — even though the
  style guide asks for "kopiér" and Danish Office uses "Sæt ind". Worth fixing
  systematically, not here.

**Conflicts:** none. I checked every changed key against my other open
PRs — no keys overlap.

**Verification:** JSON parses, key order follows en.json, placeholders and HTML
untouched. Two pre-existing trailing-space warnings disappear. Smoke-tested
against ghcr.io/euro-office/documentserver:latest (lang=da) — all 214 keys
applied and render correctly.

AI disclosure: I used Claude Opus 5 for the propagation tooling and to propose
translations. I reviewed all 214 strings myself in a sheet showing
English, German, Swedish and Danish side by side, and corrected the suggestions
where my reading differed — four of the changes above came from my own testing
of the running editor rather than from the tooling. I take responsibility for
the result.


chrip (approved):
## Summary

111 new translations and 103 corrections across 11 files, and I verified all 214 against the
base: nothing removed, no placeholder or markup mismatch, no key ordering drift, every file
parses. For a PR that touches every editor plus the mobile locales, that is a clean structural
result.

The best decision here is not translating anything new at all. Propagating strings that already
exist in Danish elsewhere, and then checking the donor before copying, is the opposite of the
usual failure mode. Catching that 26 of the 107 donors were themselves wrong, including one that
would have carried the `(intet)` placeholder into a third editor, is the whole value of doing it
that way rather than with a script.

`Find` to `Søg` and `Equation` to `Ligning` are both well argued. The `Ligning` case is the
stronger one, since English itself distinguishes the inserted object from a named formula, and
keeping `formel` for the named formulas preserves that split rather than flattening it.

| Question | Evidence | Verdict |
|----------|----------|---------|
| `markeringslinje` for `accent bar` | German renders it `Markierungsleiste`; Swedish left the entire callout string in English, so there is no third opinion to have | ✓ German is the only precedent and you followed it consistently across all 24 |
| `Billede fra lager` for `Image from storage` | German `aus dem Speicher`, Swedish `Från lagring`; your `Fra lager` matches both in sense and matches the 11 existing occurrences | ✓ keep it |
| `Kopier` without the accent | Consistency with existing occurrences is the right call for this PR | ⚠️ see below, your own #208 goes the other way |

On the accent: your reasoning is sound and I agree this is not the place to change it. The catch
is that #208 adds `Common.Views.MacrosDialog.textCopy` as `Kopiér`, so as things stand the two
PRs pull in opposite directions. Worth aligning that one key in #208 to the decision you made
here, so the systematic fix you have in mind later starts from a consistent base.

## Two leftovers worth a follow-up

**`Open Recent` now reads two ways.** Your new `Common.Views.RecentFiles.txtOpenRecent` is
`Åbn seneste`, while `PE.Views.FileMenu.btnRecentFilesCaption` still reads `Åben nylige` for the
same English source. Same menu concept, and the old one also uses the adjective `Åben` where the
imperative `Åbn` belongs, so yours is the better string of the two.

**The object-disabled sentence.** You added the clean form,
`Objektet er deaktiveret, da det redigeres af en anden bruger.` The older rendering survives
elsewhere in the same file as `Objektet er slået fra da det bliver redigeret af en anden bruger. `,
missing a comma and carrying a trailing space.

I also checked `Close`, which now maps to both `Luk` and `Afslut`. That one looks correct as it
stands, since the `Afslut` case is `PE.Views.FileMenu.btnExitCaption`, where exit rather than
close is the actual action.

## Verdict

**Approve** — a genuinely well-designed PR that verifies its own sources rather than trusting
them, with three sound terminology calls and honest flagging of the uncertain ones. The `Kopiér`
divergence with #208 is worth settling in that PR rather than this one.

bargib:
Thanks. I've updated so that all ten Danish PRs now merge cleanly with base again.

The conflicts came from #225 itself — six of the PRs touch keys it changed. The merge commits contain no new translations, only conflict resolutions, and I applied one rule throughout: whichever side holds the newer correction wins. So main won for the title-case and e-mail fixes that came in with #225, and the PR won where the PR is itself the fix — #209 and #219 correct strings that main still has wrong.

On merge order: #224 is the only one that overlaps with any other, so merging it last should avoid a cascade entirely. The other nine are independent of each other — six of them only add new keys and change nothing at all. Where #224 does overlap (#209, #212, #219), it is four keys in total and both sides already carry identical values, so those resolve themselves.

I will keep an eye on them as you go, so please do not let me hold up the rest.

If you have any suggestions on how to improve my process, feel free to share.

Cheers
Christian "Bargib" Koerner

PR #228, opened 2026-09-12
i18n(da): update the cell style gallery namings

bargib:
Brings the Danish cell style gallery in the spreadsheet editor in line with normal danish naming conventions: 60 strings in total, across the desktop and mobile locales.

## Heads-up: six names change 

This is not only a fix for untranslated text. Six styles that already had Danish names get updated ones in line with common naming:

| English | Was | Now |
|---|---|---|
| Bad | Dårlig | **Ugyldig** |
| Note | Note | **Bemærk!** |
| Calculation | `Beregner ` | **Beregning** |
| Check Cell | Tjek celle | **Kontrollér celle** |
| Linked Cell | Forbundet celler | **Sammenkædet celle** |
| Total | I alt | **Total** |

I think the change is worth it: these are the names users see in other programs, and the point of the gallery is that users recognise the styles they already know. But it is a visible change rather than a pure correction, so it should be a deliberate decision rather than something that slips through.

Two of the old values were also wrong on their own terms: `Beregner ` carried a trailing space, and `Forbundet celler` combined a singular adjective with a plural noun.

## Section headings

Six of the seven group headings had no Danish at all and showed in English above Danish style names. The seventh was written as two words where Danish needs a compound.

| English | Danish |
|---|---|
| Good, bad and neutral | God, dårlig og neutral |
| Data and model | Data og model |
| Titles and headings | Titler og overskrifter |
| Themed cell styles | Tematiserede celletypografier |
| Number format | Talformat (was "Tal format") |
| Custom | Brugerdefineret |
| No name | Uden navn |

The first five follow common Danish practice in other office programs. "Custom" and "No name" do not appear in the normal  gallery, so they follow the wording already used elsewhere in this repository.

## Mobile: the gallery was untranslated

`spreadsheeteditor/mobile` had no Danish for any cell style, although the same names already existed in the desktop locale next to it. This adds 47 strings.

The `Accent` family becomes `Farve`:

| English | Danish |
|---|---|
| Accent1 … Accent6 | Farve1 … Farve6 |
| 20% - Accent1 … | 20 % - Farve1 … |
| 40% - Accent1 … | 40 % - Farve1 … |
| 60% - Accent1 … | 60 % - Farve1 … |

Note the spacing: Danish writes a space before the percent sign, where English does not.

The remaining 23 are the named styles, taken from the desktop locale and corrected where the desktop value was one of the six above: Normal, God, Neutral, Ugyldig, Advarselstekst, Bemærk!, Beregning, Forklarende tekst, Input, Kontrollér celle, Output, Sammenkædet celle, Overskrift 1-4, Titel, Total, Komma, Komma [0], Procent, Valuta, Valuta [0].

## Known gap: the Accent styles stay English on desktop

Worth stating so it does not look like an oversight. There are no locale keys for `Accent1` through `60% - Accent6` in the desktop editor at all — those names come straight from the SDK, and only the mobile editor exposes them as `Controller.Main.SDK.*`. So after this change the mobile gallery reads "20 % - Farve1" while the desktop gallery still reads "20% - Accent1". Closing that gap needs the keys to exist upstream first; it cannot be done from the locale files.

## Verification

Checked against the cell style gallery in Danish spreasdsheet programs , and the desktop changes were confirmed in a running Euro-Office container. Both files validate as JSON, key order follows `en.json`, and the branch merges cleanly into `main` with no key overlap with any other open Danish PR.

The mobile strings could not be confirmed in the running container: the Danish mobile editor currently fails to load, because `Controller.Main.warnNoLicense` is missing from the Danish locale and `Main.jsx` dereferences it unconditionally during startup. That is an existing gap unrelated to this PR — it affects all three mobile editors and seven languages — and I will open a separate PR for it.

🤖 Generated with the help of Claude Opus 5 - all translations have been verified by myself though... :-)


chrip (approved):
## Summary

60 strings across the desktop and mobile spreadsheet locales: 53 added, 7 changed, nothing removed. No placeholder or markup mismatch, no key ordering drift, both files parse, CI green on both checks.

Flagging the six visible renames in your own description, under a "heads-up" heading, is the right way to handle this. A gallery rename is a user-visible change rather than a correction, and it should be a decision someone made rather than something that slipped through in a batch of untranslated strings. Two of the old values were also broken on their own terms — `Beregner ` carried a trailing space and `Forbundet celler` paired a singular adjective with a plural noun — so those two are unambiguous improvements regardless of the naming question.

## The one rename I would push back on

`Bad` to `Ugyldig` is the only change here where the cross-language evidence points the other way:

| English | German | Swedish | Was | Now |
|---------|--------|---------|-----|-----|
| `Bad` | `Schlecht` | `Dålig` | `Dårlig` | `Ugyldig` |
| `Good` | `Gut` | `Bra` | `God` | `God`, unchanged |

Both sibling languages translate `Bad` literally, and the old Danish `Dårlig` matched them. `Ugyldig` means invalid, which is a different axis.

The sharper problem is internal. If the rationale is that Danish Excel names this pair `Gyldig` and `Ugyldig`, then the pair should move together — and `Good` stays `God` here. As it stands the gallery reads `God` and `Ugyldig` side by side, mixing good-versus-bad with valid-versus-invalid. Either take both to the valid axis or leave both on the quality axis; the mix is harder to defend than either consistent choice.

Related, `Note` to `Bemærk!` adds an exclamation mark that appears in neither the English source nor German `Hinweis` nor Swedish `Notera`. Worth a second look, since punctuation added in translation tends to survive for years.

## Still there: three more `Tal format`

You fixed the compound in the section heading, but the same English source keeps the spaced form in three other keys:

| Key | English | Danish |
|-----|---------|--------|
| `SSE.Views.DocumentHolder.txtNumFormat` | `Number Format` | `Tal format` |
| `SSE.Views.FormatRulesEditDlg.tipNumFormat` | `Number format` | `Tal format` |
| `SSE.Views.Toolbar.tipNumFormat` | `Number format` | `Tal format` |

All three are in a file this PR already touches, so they are cheap in-place edits whenever you want them. Non-blocking.

## Verdict

**Approve** — the untranslated section headings and the two genuinely broken values are clear wins, and the renames are correctly surfaced as a decision rather than buried. I would reconsider `Ugyldig` against `God`, or move both, before merging.

bargib:
I referenced the similar naming for cells in a similar product for the naming style gallery, thus keeping it in sync with what users might be used to. I know the terminology is slightly different than the source., so a valid push back, but it was because of the context and the references I took, that I kept the translation to what is seen otherwise.

PR #227, opened 2026-09-12
i18n(da): translate most of the remaining presentation editor panels

bargib:
This adds Danish for 271 previously untranslated strings in `presentationeditor/main` and corrects 11 existing ones. Danish coverage for this editor goes from 55.8% to 66.5% (2377 of 3574 strings).

## What is translated

The new strings cover the panels that were still entirely in English: the print preview and its page setup options, the chart and shape settings dialogs, the format settings dialog, external links, the keyboard shortcut editor, the document properties and document info panels, the Advanced settings panel, and 25 smaller namespaces that had no Danish at all.

## Corrections to existing strings

These were found by running the editor in a container and walking through the UI, not by reading the files:

| Key | English | Was | Now |
|---|---|---|---|
| `Common.Translation.warnFileLockedBtnView` | Open for viewing | Åben for visning | Åbn for visning |
| `Common.Views.Comments.textOpenAgain` | Open Again | Åben igen | Åbn igen |
| `Common.Views.ReviewPopover.textOpenAgain` | Open Again | Åben igen | Åbn igen |
| `PE.Views.FileMenu.btnFileOpenCaption` | Open | Åben | Åbn |
| `PE.Views.FileMenu.btnRecentFilesCaption` | Open Recent | Åben nylige | Åbn seneste |
| `PE.Views.FileMenu.btnExitCaption` | Close | Afslut | Luk |
| `PE.Views.FileMenu.btnHistoryCaption` | Version History | Version historik | Versionshistorik |
| `PE.Views.FileMenu.btnPrintCaption` | Print | Print | Udskriv |
| `PE.Views.Toolbar.tipPrint` | Print | Print | Udskriv |
| `PE.Views.FileMenuPanels.Settings.txtAutoCorrect` | AutoCorrect options... | Autokorrektur valgmuligheder | Indstillinger for autokorrektur... |
| `PE.Views.FileMenuPanels.Settings.strPasteButton` | Show the Paste Options button when the content is pasted | Vis knappen for Indsæt-optioner når indhold indsættes | Vis knappen Indsæt-indstillinger, når indhold indsættes |

`Åben` is the adjective *open*; the imperative is `Åbn`. The Danish Style Guide maps File > Open to `Åbn`, File > Close to `Luk` and File > Exit to `Afslut` — note that `btnExitCaption` has the English text *Close*, not *Exit*, so `Luk` is the right word despite the key name. `Version historik` and `Autokorrektur valgmuligheder` were written as two words where Danish requires a compound.

## Terminology

[...]

PR #229, opened 2026-09-12
i18n(da): propagate shared Common.* strings across the editors

bargib:
Many `Common.*` strings were translated in one editor and left in English in another, even though the key and the English source text are identical. This copies those translations into the editors that were missing them: 533 insertions covering 279 distinct keys, plus ten corrections described below.

| Editor | Strings added |
|---|---|
| spreadsheeteditor/main | 191 |
| documenteditor/main | 150 |
| pdfeditor/main | 149 |
| documenteditor/mobile | 22 |
| presentationeditor/mobile | 13 |
| spreadsheeteditor/mobile | 5 |
| documenteditor/forms | 3 |

## How the candidates were chosen

A string qualifies only when the key is missing in one editor, present in another, and the Danish value is unambiguous — that is, every editor that already has the key agrees on the same Danish text. Where editors disagreed, the key was skipped rather than guessed at.

Keys that an open Danish PR already provides were also left out, so every string in this series is delivered by exactly one pull request. That excluded 144 keys, all of which belong to #227.

## Corrections rather than straight copies

Five values were wrong at the source and were corrected before being copied:

| Key | Was | Now |
|---|---|---|
| `Common.Collaboration.textComments` | Kommentare | Kommentarer |
| `Common.define.chartData.textLine` | Linie | Linje |
| `Common.VersionHistory.txtErrorLoadHistory` | Indlæsningshistorik fejlede | Indlæsning af historik fejlede |
| `Common.Controllers.History.txtErrorLoadHistory` | Fejl ved indlæsningen af historik | Indlæsning af historik fejlede |
| `Common.Utils.ThemeColor.txtaccent` | Accent | Fremhævning |

`Kommentare` and `Linie` are German that had found its way into the Danish files. The two `txtErrorLoadHistory` keys have near-identical English source text and now read the same in Danish. `Accent` is the theme colour in the palette tooltip, which `utils.js` composes as "Fremhævning 1" through "Fremhævning 6" — the name other Danish office programs use for these theme colours.

Two strings in the password dialog are also corrected, in the four editors where they appear:

| Key | Was | Now |
|---|---|---|
| `Common.Views.PasswordDialog.txtDescription` | Indstil et kodeord for at beskytte dette dokument | Angiv et kodeord for at beskytte dette dokument |
| `Common.Views.PasswordDialog.txtIncorrectPwd` | Bekræftelsesadgangskode er ikke identisk | Bekræftelse af kodeordet er ikke identisk |

`Indstil` is the wrong verb for setting a value the user types; Danish uses `Angiv`, as Swedish does with `Ange`. The second replaced a heavy compound noun with the construction Swedish uses, `Bekräftelse av lösenordet`.

## A known, deliberate inconsistency

The second commit brings four existing occurrences into line with the corrected values, so the same key does not say two different things. Two further occurrences, both in `presentationeditor/main`, are deliberately left alone: changing them would create a merge conflict with #227. They can follow once that lands.

Two more will resolve themselves when other open PRs merge — `textComments` in `spreadsheeteditor/mobile` via #209, and `chartData.textLine` via #224.

## Verification

Every Danish file still parses, key order follows `en.json` in all eight, and no placeholder values were introduced. The branch merges cleanly into `main`, and the changes were applied to a running Euro-Office container and checked in the UI.

On the word "Accent": it needs four different Danish words in this codebase depending on context, and the repository already makes the distinction in places. `Accenter` for the diacritical marks in the equation editor, `markeringslinje` for the accent bar in callout shapes, `Markeringsproces` for the SmartArt layout, and `Fremhævning` for the theme colour introduced here.

## AI
I used Claude Code with Opus 5 to assist in this work.


chrip (approved):
## Summary

533 insertions across eight files plus ten corrections, propagating `Common.*` strings that were translated in one editor and left in English in another. I verified against the merge base: 533 added, 10 changed, nothing removed, no placeholder or markup mismatch, no key ordering drift, all eight files parse. CI green on both checks.

Two of your numbers I checked rather than assumed, and both are exact. The 533 insertions do cover **279 distinct key names**, the difference being keys that land in more than one editor. And this branch shares **zero** keys with #227, which is precisely what your "every string delivered by exactly one pull request" rule promises — I compared all 282 keys in #227 against all 543 here and the intersection is empty.

The selection rule is the best thing in this PR. Requiring that every editor already holding the key agrees on the same Danish, and skipping the key otherwise rather than picking a winner, is what keeps a propagation pass from quietly spreading one editor's mistake into three more. The five corrections-before-copying show the rule doing real work, particularly the two `txtErrorLoadHistory` keys that had near-identical English and two unrelated Danish renderings.

`Accent` to `Fremhævning` is well argued. Checking how `utils.js` composes the palette tooltip into `Fremhævning 1` through `Fremhævning 6`, rather than translating the bare word in isolation, is the kind of thing that only shows up if you look at the consuming code.

## What this leaves behind

Propagation necessarily lands a correct value next to an older, different one. None of these are defects in this PR and all involve keys it did not touch, but they are what a later sweep inherits:

| English | Propagated value | Still elsewhere |
|---------|------------------|-----------------|
| `Print` | `Udskriv` | `Print` |
| `Open` | `Åbn` | `Åben` |
| `Version History` | `Versionshistorik` | `Version historik` |
| `Number format` | `Talformat` | `Tal format` |
| `General` | `Standard` | `Generel`, `General ` |
| `Line` (pdfeditor) | `Linje` | `Linie` |

Several of these are already being settled elsewhere in the series: `Print` and `Open` by #227 in presentationeditor, `Tal format` partly by #228, `Linie` by #224's sweep. The `Redo` splits are #207's. So the list shrinks on its own as the series lands — worth a sentence in the description saying so, since a reviewer seeing `Udskriv` and `Print` in one file will otherwise read it as an oversight.

`General` to `Standard` is the one I would double-check. Three Danish values now coexist for that source, and `Standard` is semantically distinct from `Generel` rather than a spelling variant, so it is a terminology decision rather than a propagation.

## Merge safety

No key anywhere in this branch is set to a different value by any of the fourteen other open Danish PRs. Merge order is irrelevant.

## Verdict

**Approve** — a well-designed propagation pass with a selection rule that refuses to guess, numbers that check out exactly, and honest correction of the five donors that were wrong. `General` to `Standard` is worth one more look.

PR #230, opened 2026-09-12
fix(i18n-da): add the mobile strings whose absence crashes the editor

bargib:
The Danish mobile editors currently fail to load. This adds the 25 locale strings whose absence causes it.

## What happens

The mobile controllers read some translations straight off the locale object and call String methods on the result, without checking that the value exists:

```js
const warnNoLicense = _t.warnNoLicense.replace(/%1/g, __COMPANY_NAME__);
```

When the key is missing from the locale, the value is `undefined` and `.replace` throws. Three of these run unconditionally during startup, in `applyLicense`, so the editor never finishes loading: the document does not appear and the user sees "an error has occurred while working on the document" instead.

```
Uncaught TypeError: Cannot read properties of undefined (reading 'replace')
    at applyLicense (Main.jsx:867:48)
    at onDocumentContentReady (Main.jsx:771:14)
```

Danish is not alone. Measured against `main` in this repository, each mobile editor has 45 locale files, and ten of them lack `warnNoLicense` in at least one editor:

| Editor | Locale files | Lack `warnNoLicense` |
|--------|--------------|----------------------|
| documenteditor | 45 | 8 — be, bg, da, fi, no, pl, sl, sv |
| presentationeditor | 45 | 9 — the same, plus uk |
| spreadsheeteditor | 45 | 10 — the same, plus uk and vi |

I have only checked Danish in a running editor, but the same three lines run for every locale, so the same crash is likely in the others. Swedish is worth noting in particular, since this project uses it as a translation reference.

An earlier version of this description gave 43 files and named seven locales, measured in a 9.3.1 container rather than against this repository. The table above replaces it.

## What this adds

25 strings across the three mobile editors, found by scanning the source for this dereference pattern and checking each key against the Danish locale:

| Editor | Strings |
|---|---|
| documenteditor | 11 |
| presentationeditor | 8 |
| spreadsheeteditor | 6 |

Eight are the license strings that break startup. The other 17 are in `Error.jsx`, where the error handler itself would throw while trying to report an error — latent rather than immediate, but the same bug.

After this change, no unguarded dereference in the mobile source points at a key the Danish locale lacks.

## Wording

Taken from the desktop locale of the same editor wherever the key exists there, with two adjustments. The mobile source has no `<br>` in `warnNoLicense`, and no full stop at the end of the three `split*` strings, so the Danish follows the mobile source rather than the desktop one.

The four `errorInconsistentExt*` strings have no desktop equivalent and are new, following the structure German and Swedish use.

One existing value is corrected: `textNoLicenseTitle` in the spreadsheet editor had a trailing full stop that neither the English source nor the other two editors have.

## This fixes the symptom, not the cause

The unguarded dereferences are still there. The next locale that is missing one of these keys will break in exactly the same way, and a translation PR cannot prevent that — it needs a fallback in `Main.jsx` and `Error.jsx`. I am flagging it rather than fixing it, since it is a code change rather than a locale one.

## Danish mobile coverage is still low

This PR is about making the editors load, not about completing the translation. A large number of mobile strings are still in English:

| Editor | Danish | Missing |
|---|---|---|
| documenteditor/mobile | 115 of 812 (14%) | 697 |
| presentationeditor/mobile | 223 of 546 (41%) | 323 |
| spreadsheeteditor/mobile | 495 of 867 (57%) | 372 |

Those will follow in a later PR. Keeping them out of this one means this change stays small enough to review as the bug fix it is.

## Verification

Applied to a running Euro-Office container and loaded all three mobile editors with `lang=da`. They now load normally and the TypeError is gone from the console. All three locale files parse and keep their key order from `en.json`.

## AI
Used Claude Opus 5 to assist in making this fix.


chrip (approved):
## Summary

This is a bug fix rather than a translation PR and it should be treated as the most urgent thing in the series. The Danish mobile editors fail to load, and this adds the 25 locale strings whose absence causes it.

The diagnosis is right and well evidenced. The controllers dereference locale values directly, `_t.warnNoLicense.replace(...)` throws when the key is missing, and three of those run unconditionally in `applyLicense` during startup, so the editor never finishes loading. Including the actual stack trace and the call path through `onDocumentContentReady` makes this verifiable rather than assertable.

Structurally it is clean on every check: 25 added, 1 changed, nothing removed, every `%1` placeholder and every `<br>` preserved between the English source and the Danish, no key ordering drift, all three files parse. CI green on both checks.

I checked the completeness claim too. For every one of the eleven affected key names, all three mobile editors now have Danish where the English source exists — the three `split*` keys are genuinely absent from the spreadsheet editor's English file rather than skipped. The single changed value, `textNoLicenseTitle` losing a trailing full stop, brings the spreadsheet editor in line with the other two and with the English source.

## One correction: the affected-locale list is off

This matters because you use it to argue other locales are likely broken too, and someone will act on that list.

Measured against `main` in this repository, using `apps/documenteditor/mobile/locale`:

| | Your description | Measured here |
|---|---|---|
| Locale files | 43 | 45 |
| Have `warnNoLicense` | 35 | 37 |
| Lack it | 8 | 8 |
| Which ones | be, bg, da, fi, pl, uk, vi | be, bg, da, fi, **no**, pl, **sl**, **sv** |

The count of eight is right and the description only names seven, so one was already missing. More importantly three names differ: Norwegian, Slovenian and Swedish are affected and are not on your list, while Ukrainian and Vietnamese are not affected but are.

The honest caveat is that you measured a 9.3.1 container and I measured this repository's `main`, and those can legitimately differ. So this is a "please reconcile" rather than a flat contradiction. Either way, `no`, `sl` and `sv` are worth re-checking before anyone relies on the list, and Swedish in particular is a locale this project actively uses as a translation reference.

## Verdict

**Approve, and merge this ahead of the rest.** Every other PR in the series improves Danish; this one makes the Danish mobile editors start at all. The diagnosis, the placeholder handling and the cross-editor completeness all check out. Only the affected-locale list needs reconciling, and that is a description edit rather than a code change.

bargib:
@moodyjmz @chrip — a request on sequencing.

This one is a crash fix rather than a translation: the Danish mobile editors do not load at all, because three controllers dereference locale values during startup and throw on the missing keys. chrip's review puts it as "Approve, and merge this ahead of the rest."

So please take this one first. But I would be grateful to see my other fourteen translation PRs land as well. All fifteen are approved, mergeable and green, and the review confirmed that no key is set to different values by two branches, so they can go in any order. The open queue is what now holds me up — further Danish work stacks on branches that have not landed.

To bring Danish into full working order I need these merged, so I can apply what I have learned here, and the terminology established along the way, to the other editors and the desktop apps.

PR #231, opened 2026-09-13
i18n(da): Danish for the three mobile editors

bargib:
## Danish for the three mobile editors

The mobile editors were the least translated part of the Danish locale. This
brings them to the same level as the rest.

| | before | after |
|---|---|---|
| documenteditor/mobile | 12.8% | 95.9% |
| presentationeditor/mobile | 39.4% | 96.2% |
| spreadsheeteditor/mobile | 56.4% | 93.3% |

The work was split by what each string needed, and each part is its own commit.

**Strings that already had Danish elsewhere** (623 keys). The same English
source was already translated in another editor or in the desktop files, so the
existing wording was carried over rather than a new one invented. A further 58
keys had Danish that contradicted a decision made in this series and were
brought into line.

**Strings with competing Danish translations** (173 keys, 76 texts). The same
English source already had two or more different Danish translations in the
repository, so propagating any one of them would have spread a choice nobody
had made. Each was decided against the English source, with German as the main
reference and Swedish as a secondary one. Where those left real doubt, the
other locale files in the same directory were counted — there are 45 of them,
and a majority across languages is a better signal than one language alone.
That settled `Tab` as Ark, `Markup` as Ændringer and `Delay` as Forsinkelse,
in all three cases against German, which had kept the English word.

**Strings with no Danish anywhere** (509 keys, 332 texts). Translated from the
English source and checked against the Danish style guide. Its section on
compounds describes the exact construction behind *The destination cells range
can contain data*: English noun stacks become relative or prepositional
clauses in Danish rather than a new compound noun, so that message now names
the cells the user is moving to instead of inventing a term for them.

Where a language other than German had understood the source better, that
reading was followed. *Line size* is the stroke's thickness, not its size —
Russian, Korean and Chinese all say so, while German and the Romance languages
calque the English word.

### Consequence changes

Every proposal was checked against the Danish that already existed elsewhere
in the repository. Fifty-two differed; in five the existing translation was the
better one and was adopted, including `No Matches` as Ingen resultater, which
four editors already used.

Thirty-five existing strings are corrected because the new ones sit beside
them and would otherwise read inconsistently:

- Four of the six page-number positions were missing while *Center Top* and
  *Center Bund* stood untranslated. The set now reads Øverst and Nederst
  throughout.
- The two leave-page buttons and the dialog that quotes them now use the same
  words.
- `Through Black` was left in English by the first commit in this branch. The
  desktop files have had *Gennem sort* all along, and 42 of the 43 languages
  translate it.

The 112 keys still missing from these files are delivered by the other open
pull requests in this series and were left out here on purpose, so the two
sets do not conflict.

Terminology follows the Danish style guide and common usage in comparable
office software. As always with translation, the wording is a judgement call in
places, and review is welcome.

## AI


chrip (approved):
## Summary

The largest PR in the series by a wide margin: 1,305 added and 93 changed across the three mobile locales, taking documenteditor/mobile from 12.8% to 95.9%. I verified against the merge base — nothing removed, no placeholder or markup mismatch anywhere in the 1,398 touched keys, no key ordering drift, all three files parse. CI green on both checks.

Splitting by *what each string needed* rather than by file is the right decomposition for work at this size, and it is what makes 1,398 keys reviewable at all. The middle category is the one that shows judgment: recognising that 173 keys had competing Danish translations already in the repository, and that propagating any one of them would spread a choice nobody had actually made, is a distinction most contributors would not draw.

Counting the other 45 locale files to break a tie, and then following that majority against German where German had kept the English word, is a genuinely good method. It is also falsifiable, which is why I could check it.

## The trailing spaces are correct

My whitespace check flagged seven strings, and I want to record that all seven are right, because they look like defects and a later reviewer will flag them again:

| Key | English source | Danish |
|-----|----------------|--------|
| `Common.Collaboration.textLineSpacing` | `Line spacing: ` | `Linjeafstand: ` |
| `Common.Collaboration.textNot` | `Not ` | `Ikke ` |
| `Edit.textInvalidName` | `The file name cannot contain any of the following characters: ` | `Filnavnet må ikke indeholde følgende tegn: ` |
| `Main.SDK. -Section ` | ` -Section ` | ` -Sektion ` |
| `Main.SDK.Odd Page ` | `Odd Page ` | `Ulige side ` |
| `View.Edit.textInvalidName` (presentation) | same as above | same as above |
| `View.Edit.textInvalidName` (spreadsheet) | same as above | same as above |

Every one matches its English source's leading and trailing whitespace exactly. These are strings the UI concatenates, so the spaces are load-bearing and stripping them would be the bug.

## Small divergences worth a look

None of these block, and some may be deliberate:

| English | Values now in the same file | Note |
|---------|------------------------------|------|
| `Tab` (spreadsheet mobile) | `Ark` on `Controller.Main.SDK.Tab`, `Fane` on `View.Settings.txtTab` | your `Ark` decision is sound for the sheet sense; worth confirming the settings key is the other sense |
| `Shape` (presentation mobile) | `Figur`, `Form` | |
| `Anonymous` (spreadsheet mobile) | `Anonym`, `Anonymt` | inflection, likely context-driven |
| `Auto` (document mobile) | `Auto`, `auto` | capitalisation only |
| `Time` (spreadsheet mobile) | `Klokkeslæt`, `Tid` | you settled this one deliberately |

The `Tab` row is the one I would confirm rather than change. `Ark` is clearly right where `Tab` means a sheet, and if `View.Settings.txtTab` is the tab-character sense then `Fane` is wrong there for the reason #212 established — but if it is a settings tab, `Fane` is right and there is nothing to do.

## A note for #209

This PR sets `View.Edit.textJustified` to `Lige margener` in the spreadsheet mobile file, which fixes one of the five `berettiget` occurrences I raised on #209, with exactly the value I suggested. The sibling `Common.Collaboration.textJustify` in `documenteditor/mobile` still reads `Juster berettiget`, and since this PR is already open in that file, it would be a natural one-line addition here.

## Merge safety

This branch shares two keys with #212, both with identical values. Across all fifteen open Danish PRs there is no key that two branches set to different values, so merge order is irrelevant throughout the series.

## Verdict

**Approve** — 1,398 keys with a decomposition that makes them checkable, a tie-breaking method that is both sensible and verifiable, and correct handling of the whitespace-significant strings that this kind of bulk work usually damages. The divergences above are follow-ups, not blockers.

PR #233, opened 2026-09-17
i18n(da): give each button in the equation gallery its own name

bargib:
> **Builds on #232.** This branch is cut from that one, so until #232 merges
> GitHub shows both sets of changes here. The 353 belonging to this pull
> request are the five commits after "name the set operators after what they
> do". Eleven keys are touched by both — capitalisations in #232 that these
> corrections then rewrite — which is why the two are stacked rather than
> parallel: as separate branches they would conflict, and every string in the
> series has so far been delivered by exactly one pull request.

## The equation gallery said the same thing about different buttons

120 distinct buttons in the equation gallery shared 65 Danish tooltips. Hover
over eighteen different brackets and all eighteen said "Enkelt klamme"; the
five variants of each integral and each large operator were likewise
indistinguishable. After this they have 119 texts between them, one per
button but for a single pair that genuinely means the same thing.

This is not the same fault as the pull requests before it. Those corrected
Danish that was wrong. Here most of the Danish is defensible in isolation —
what is missing is the distinction the source draws, which German keeps and
Danish had lost.

### Corrections that are not about detail

A few turned out to be plain errors, found while separating the rest.

**A 2x2 matrix labelled 3x3.** The template for a 2x2 identity matrix with
blank off-diagonal cells said 3x3 in Danish. Picking by the tooltip inserted
the wrong template.

**A differential read as "different".** `Differential x` had become
"Forskelligt x". German has Differenzial.

**A contour integral turned inside out.** `Contour integral` had become
"Integreret kontur" — a contour that has been integrated, rather than an
integral along one.

**A co-product read as a partial product.** "Delprodukt" is a partial
product; the co-product is the dual construction to the product.

**An accent named after what its name means.** `Acute` had become "Skarp",
because aigu is French for sharp. The character is the accent aigu. `Grave`
had become "Grav", a hole in the ground.

### Where the reference languages could not be used

Swedish and Dutch have the identical fault in the bracket gallery — one word
for every family, another for every single character — so neither could serve
as a check. German separates the families correctly but errs on the singles,
rendering "Right curly bracket" as a left round one. Those were verified
against Russian, French and Italian.

For the accents, all four reference languages read `Check` as the verb —
prüfen, markera, vérifier — where the character is the caron.

### Terminology

The mathematical terms are checked against Danish sources (universities online material) rather than
translated by ear:

| | |
|---|---|
| bracket families | Lex, Danish Wikipedia |
| floor and ceiling | the spreadsheet functions GULV and LOFT |
| the vertical separator | read "given" in P(A\|B), "where" in set-builder notation |
| integral types | DTU's material on integration in several variables |
| intersection, union, surface integral, identity matrix | the SDU English-Danish mathematics dictionary |
| sparse matrix | DTU's English-Danish vocabulary: tyndt besat matrix |
| the diacritics | Lex and the Danish Wikipedia article on diacritical marks |

Two are deliberate departures, both noted so a reviewer can see they were
decided rather than missed. `Contour integral` is konturintegral where the
SDU dictionary gives kurveintegral, because the symbol here is the closed
one. The breve is **Bue** rather than the orthographically correct brevis,
because the gallery is chosen by eye and few readers know the word — Swedish
takes the same route with Kort.

`Co-product` is the one term with no Danish source at all. Neither SDU, DTU,
Aalborg, Aarhus, the Danish Wikipedia nor Swedish and Norwegian material has
it. Coprodukt is a judgement, and it is marked as such.

## AI

Every one of the 353 strings was reviewed and approved by me individually
the English source, the German and Swedish translations, the current Danish,
the proposal, and the source each term was checked against. Several proposals
were sent back: the bracket terminology was reworked twice, and two accent
names were replaced with simpler ones after I asked whether the specialist
terms were needed. Just so you understand that this more than an AI translation ;-)


juliusknorr (approved):

chrip (approved):
**Approve** — no blockers.

Verified against the merge base: **0 keys added, 379 changed, 0 removed** across 4 file(s). All locale files parse, no placeholder or `<br>` mismatch against the English source, no key-order drift, and every added key exists in `en.json`. Cross-checked against the other 33 open Danish PRs: no key is set to a conflicting value by another branch, so merge order does not matter. CI green, DCO signed.

PR #234, opened 2026-09-17
i18n(da): translate the keyboard shortcut descriptions across the editors

bargib:
## The shortcut list was in English in three of the four editors

`Common.Controllers.Shortcuts.*` is shared: the same keys, with the same
English source, appear in all four editors. Danish had them in the document
editor and almost nowhere else, so anyone opening **File → Advanced settings
→ Keyboard shortcuts → Customize** in the PDF editor got 128 descriptions in
English.

| | |
|---|---|
| pdfeditor | 128 |
| presentationeditor | 60 |
| spreadsheeteditor | 38 |
| documenteditor | 1 |

## They were read before they were propagated

Copying 227 strings is a five-minute job. Reading them first is not, and it
is where this pull request spent its time: 28 of the existing Danish texts
were rewritten before anything was copied. A rewrite before propagation costs
one change; after it costs four.

Six were wrong rather than merely heavy.

**A header at the bottom of the page.** "Move to the lower header" had become
"Gå til det nederste sidehoved" — the bottommost header, which is a
contradiction in terms. The shortcut is Page Down, it only works with the
cursor already inside a header or footer, and *lower* and *upper* are
therefore a direction through the document, not a position. It should have
read "the next". Every reference language translated it literally: untere
Kopfzeile, intestazione inferiore, encabezado inferior, en-tête inférieur.
Russian even transliterated *header* and *footer* into the sentence while
using the correct Russian word in the parenthesis beside it.

**A paragraph switched between a noun and an adjective.** "Switch a paragraph
between justified and left-aligned" mixed the two where the two sibling
strings use adjectives on both sides.

**Content that contained content.** "If it does not contain content, content
is created."

**A display that was shown.** "Show or hide the display of nonprinting
characters" said show and display for one thing.

**A key that was pressed, but not pressed on.** "Tryk den ønskede genvej"
needs the preposition in Danish.

**The only curly quotation marks in the repository.** Danish takes the
straight ones. Every Danish locale file was checked; this was the one string.

The remaining 22 were legible but heavy, and mostly the same fault: English
noun stacks left standing. "Markørvalgtilstand" is one twenty-one-letter word
where German can write Markierungsauswahlmodus and Danish cannot.

Two carried an inherited misspelling — *liste-* written as *lister-* — which
propagation would have given three new homes. That is the third time in this
series that propagation would have multiplied an existing error, and the
reason the reading happens first.

## Where the interface was consulted rather than the dictionary

The descriptions name panels and menus by their labels, so the labels were
looked up in the strings the editors actually use rather than translated
again: Find and Replace is **Søg og erstat**, Comments is **Kommentarer**,
Navigation is **Navigering**.

Two terms are worth noting because they look like candidates for change and
are not:

**Column break** stays *spalteskift*. The key lives in the document and PDF
editors, where a column is a text column, not a spreadsheet column. German
says Spaltenumbruch.

**Desktop Editors** and **Online Editors** stay *skrivebordseditorerne* and
*onlineeditorerne*. They read like product names in the source, but four of
the five reference languages render them as ordinary descriptive words, and
the fifth does so for one and not the other in the same sentence. Four keys
in this family already use the Danish words, and *editor* is the word the
rest of the interface uses: Dokumenteditor, Regnearkseditor, mobileditorerne.

## Not translated

287 keys whose English "translation" is the key name itself —
`txtLabelMoveToLeftChar` reads "MoveToLeftChar". No language translates
those. They are identifiers, not text.

## AI

Every one of the 265 strings was reviewed and approved by me individually
the English source, the German and Swedish translations, the current Danish,
the proposal, and for each rewrite a note saying why. Two proposals were sent
back: my first fix for the header/footer strings removed the article to dodge
a gender clash and left the actual error standing, and a later one used the
definite form where the bare form reads better.


chrip (approved):
**Approve** — no blockers.

Verified against the merge base: **227 keys added, 38 changed, 0 removed** across 4 file(s). All locale files parse, no placeholder or `<br>` mismatch against the English source, no key-order drift, and every added key exists in `en.json`. Cross-checked against the other 33 open Danish PRs: no key is set to a conflicting value by another branch, so merge order does not matter. CI green, DCO signed.

PR #235, opened 2026-09-17
i18n(da): translate the header and footer tab

bargib:
## A whole tab in English

The Header & Footer tab had no Danish at all — not one of its buttons,
checkboxes or tooltips. It also appears in the PDF editor, which borrows the
document editor's ribbon when editing, so it was two editors showing the same
untranslated tab.

| | |
|---|---|
| documenteditor | 37 |
| presentationeditor | 8 |
| spreadsheeteditor | 2 |
| pdfeditor | 1 |

## Written, not copied

Most of these sources already had Danish somewhere in the repository, so
propagating them would have been quick. Reading what was there first is why
it was not done that way.

**Edit header** and **Remove header** read *Rediger overskrift* and *Fjern
overskrift*. An overskrift is a heading — the thing that starts a section.
A header is the band at the top of every page. The two are different parts
of a document, and the shortcut list in the same editor uses *sidehoved*
correctly.

**Edit header or footer** read *Rediger header eller sidefod* — half
translated, with the English word left standing next to its Danish sibling.

Copying either would have given them new homes.

## Where Danish already existed, it was used

**Link to previous** is *Link til forrige*. The same checkbox is already
spelled that way in the right-hand panel and in the mobile editor. My first
draft invented a third wording for one control; the interface settled it.

**To current position** and **Insert current date and time** likewise follow
the strings already in the files rather than being translated again.

## Two choices a reviewer may want to weigh

**Different odd and even pages** is *Forskellige lige og ulige sider*.
Danish names the even before the odd; English does the opposite.

**Date & Time** is *Dato og tid*, not *Dato og klokkeslæt*, even though
klokkeslæt is the precise word for a time of day and German uses its
equivalent (Uhrzeit). All nine occurrences of this source in the files say
*tid*, and Swedish says *tid* as well. Changing it is a change to all nine,
not to one. Worth noting that *Klokkeslæt* is right where it is used — the
four keys where Time is a number-format category.

## Three corrections found by reading

None of these is in the tab. They were found by looking up what the files
already say for the sources the tab uses.

**A missing letter.** *Sidehoved- og sidefodsinstillinger* should be
*indstillinger*. It is the only occurrence in the repository. It is corrected
to *Indstillinger for sidehoved og sidefod*, which is what the presentation
editor already says for the same English source.

**A marker in two voices.** The text the editor shows beside a linked header
reads *Magen til tidligere* on the desktop and *Samme som forrige* on mobile.
The first is colloquial, and *tidligere* is not *forrige*. Every reference
language says same as previous: Dasselbe wie zuvor, Samma som föregående,
Zelfde als vorige, Identique au précédent. The desktop now follows the
mobile, so the checkbox says *Link til forrige* and the marker says *Samme
som forrige*, with one word for previous rather than two.

**A job where a placement was meant.** Position had become *Stilling* in one
key, which in Danish is a job or a bodily posture. The other seven
occurrences of Position say *Placering*.

## AI

Every one of the 52 strings was reviewed and approved by me individually
the English source, the German and Swedish translations, the current Danish
where there was any, the proposal, and a note for each correction saying why.
Four were sent back: Link to previous, which had to match the existing
checkbox rather than introduce a third wording; Position, which I asked to be
checked against Placering rather than Lokation; Date & Time, which I asked
about; and a tooltip that had drifted from the wording already in the files.


chrip (approved):
**Approve** — no blockers.

Verified against the merge base: **48 keys added, 4 changed, 0 removed** across 4 file(s). All locale files parse, no placeholder or `<br>` mismatch against the English source, no key-order drift, and every added key exists in `en.json`. Cross-checked against the other 33 open Danish PRs: no key is set to a conflicting value by another branch, so merge order does not matter. CI green, DCO signed.

PR #236, opened 2026-09-17
i18n(da): finish the quick access strings and name the flowchart storage symbols

bargib:
Two small groups, one commit each. They have nothing to do with each other
beyond both being too small to carry a pull request alone.

## The button that hid the shortcut dialog

Smoke-testing #234 turned up something the translation itself could not fix:
the keyboard shortcut list could not be found. The row in **File → Advanced
settings** is headed *Tastaturgenveje*, but the button that opens it read
**Customize**, so nothing about it suggested it had anything to do with
shortcuts.

Looking that up turned out to be a whole family. The document editor has it
complete in Danish; the other three editors each hold a different subset.

| | |
|---|---|
| Customize | Tilpas |
| Quick access toolbar | Værktøjslinjen Hurtig adgang |
| Customize Quick Access Toolbar | Tilpas værktøjslinjen Hurtig adgang |
| Customize quick access | Tilpas Hurtig adgang |

None of that Danish is new. It is copied from wherever in the repository it
already stood, into the thirteen places that lacked it.

Five existing strings are corrected on the way past. The bar is called
*Hurtig adgang*, and that is a name. The style guide puts names in capitals
and the common noun in front of them in lowercase, the way Danish writes
*fanen Hjem* or *panelet Kommentarer*. Three of the four strings did that
already; *Tilpas hurtig adgang* did not, in five places.

The Visio editor lacks these too, but it has no `da.json` at all, so it
cannot be given them here.

## Storage that was being kept rather than stored

Three shapes in the flowchart gallery, across four editors. Two things were
wrong with them.

**The wrong word.** *Opbevaring* is not what this repository calls storage.
Everywhere else it is *lager* — Fra lager, Billede fra lager, Dokument fra
lager, twenty-one strings in all.

**Two that did not hold together.** *Direkte adgang opbevaring* is two nouns
set side by side with nothing joining them, which is the English pattern and
not the Danish one. *Opbevaring af sekventiel adgang* turns it inside out: it
is not the access that is being stored.

| | |
|---|---|
| Flowchart: Internal storage | Flowdiagram: Internt lager |
| Flowchart: Direct access storage | Flowdiagram: Lager med direkte adgang |
| Flowchart: Sequential access storage | Flowdiagram: Lager med sekventiel adgang |

Swedish solves it with *lagring* — Intern lagring, Direktansluten lagring,
Sekvensiell lagring — and German separates Zentralspeicher from Datenträger
mit direktem and mit sequenziellem Zugriff. Danish takes *lager* rather than
*lagring*, because the shape is the place and not the act, and because
*lager* is the word the rest of the files use.

The PDF editor had none of the three. They are added in the same pass.

## A note for whoever touches the spreadsheet editor next

Two of these strings cannot reach the screen in any language, and it is worth
recording why rather than leaving the next person to find it again.

In the spreadsheet editor the view that owns `txtCustomize` and
`strKeyboardShortcuts` is `SSE.Views.FileMenuPanels.MainSettingsGeneral`,
but their keys in the locale files sit under
`SSE.Views.FileMenuPanels.Settings`. The lookup never matches, so the editor
falls back to the English defaults in the JavaScript. The other three editors
name that view `Settings`, and there the same keys work.

`txtCustomizeQuickAccess` is defined in the same class and its key is
correctly under `MainSettingsGeneral` — which is why it appears in Danish
while the two beside it do not.

This is a source-side naming mismatch, not a translation gap. The Danish is
in place and will appear the day it is corrected. Nothing here tries to work
around it.

## AI

Every one of the 30 strings was reviewed and approved by me individually
the English source, the German and Swedish translations, the current Danish
where there was any, the proposal, and a note for each correction saying why.
The capitalisation of *Hurtig adgang* was my catch on reading the sheet: the
first draft left the name in lowercase, and correcting it brought five
existing strings into the batch that were not in it before.


chrip (approved):
**Approve** — no blockers.

Verified against the merge base: **16 keys added, 14 changed, 0 removed** across 4 file(s). All locale files parse, no placeholder or `<br>` mismatch against the English source, no key-order drift, and every added key exists in `en.json`. Cross-checked against the other 33 open Danish PRs: no key is set to a conflicting value by another branch, so merge order does not matter. CI green, DCO signed.

PR #237, opened 2026-09-18
i18n(da): translate the document editor's toolbar and context menus

bargib:
Two hundred and fifty-six strings in the toolbar, the context menus, the
View tab and the navigation panel. The first of three pull requests that
finish the Danish document editor.

## Inside is not Insert

Three strings were not clumsy Danish but a different command:

```
Inside borders               Indsæt rammer
Inside horizontal borders    Indsæt vandrette rammer
Inside vertical borders      Indsæt lodrette rammer
```

Someone read *Inside* as *Insert*. A user following the Danish would
press the wrong button.

## Borders are kanter

Established Danish terminology names these edges one by one, and this menu is the same menu, so it now uses the same words:

```
Bottom borders               Nederste kant
All borders                  Alle kanter
Outside borders              Ydre kanter
Inside horizontal borders    Indvendig vandret kant
```

Note the change of number: singular for the individual edges, plural
for *Alle kanter* and *Ydre kanter*. The source is plural throughout;
the Danish follows established usage rather than the source.

One hundred and thirty-eight strings elsewhere still say *ramme*. They
are a correction of their own and not in this pull request.

## Smaller things worth naming

| | |
|---|---|
| `Bullet` | read **Kugle** — the projectile. A bullet is a punkttegn. |
| `Scripts` | read **Manuskripter**. It is the group holding superscript and subscript in the equation bar, not a film script. |
| `Merge Shapes` | had no Danish at all. The five operations now use the established Danish names for these five operations: Foren, Kombiner, Fragmentér, Kryds, Træk fra. |
| `Alpha`, `Iota`, `Omicron`, `Upsilon` | take Danish spellings: alfa, jota, omikron, ypsilon. The other seventeen Greek letters are spelt the same in both languages and are unchanged. |

German would have led the way here rather than helping. For *Add space
after paragraph* it says **Leerzeichen**, which is the space character
between two words; the string is about the distance between paragraphs.
Swedish has the right word.

### Consequence change

One string reading "Bund" becomes "Nederst", because the context menu's
Top now reads Øverst and the pair has to match.

## AI

each change, before any of it was written. Several of the decisions above came
from checking established Danish terminology rather than from the
style guide — the edge names, the shape merge names and the spelling of
the Greek letters were all settled that way.


chrip (approved):
**Approve** — no blockers.

Verified against the merge base: **235 keys added, 20 changed, 0 removed** across 1 file(s). All locale files parse, no placeholder or `<br>` mismatch against the English source, no key-order drift, and every added key exists in `en.json`. Cross-checked against the other 33 open Danish PRs: no key is set to a conflicting value by another branch, so merge order does not matter. CI green, DCO signed.

PR #238, opened 2026-09-18
i18n(da): translate the document editor's chart, print and settings dialogs

bargib:
Two hundred and seventy strings across the chart axis dialog, the print
pane, the paragraph and list dialogs, the file menu and the advanced
settings. The second of three pull requests that finish the Danish
document editor.

## A thousandfold error waiting to happen

English *billion* is Danish **milliard**, and English *trillion* is
Danish **billion**. Translate them by their looks and every number on the
axis is a thousand times too large.

The existing Danish had the sense right and the spelling wrong —
*Bilioner* with one l. German has it right too. It is worth knowing that
the trap is there, because the next language to be filled in will meet
it.

## Three that were read rather than translated

| | |
|---|---|
| `Between Tick Marks` | read **Mellem flueben**. A flueben is a checkmark. These are the marks along an axis. |
| `Next to axis` | read **Næste til aksen**, as though *Next* meant next in a sequence. It means beside. |
| `Major` / `Minor` | read **Stor** and **Lille**. On an axis they are primary and secondary, which is what German says. |

`Data Labels Separator` read *Data etiket separatore*: written as three
words, and *separatore* is not a Danish word.

## Positions

Chart placements and margins now read **Øverst** and **Nederst** rather
than *Top* and *Bund*, which is the established Danish wording for a
position in a dialog.

One label needed a word English does not have. In the cell text
alignment menu the vertical and horizontal options sit in the same list,
separated by a rule. Elsewhere the two axes are named in a single
phrase — *Juster centreret til venstre* — so both can be called
*centreret*. Here, two identical items would stand three rows apart, so
the vertical one reads **Centreret lodret**.

## The line style names could not be guessed

`Miter` is **Mitra** — a bishop's mitre. It is the established Danish
name, alongside Rund and Facet, and it is not a word anyone would
arrive at by reasoning.

`Portrait` and `Landscape` are **Stående** and **Liggende**. A *landskab*
is a painting.

### Consequence changes

Seven strings reading "Bund" become "Nederst". They are not new
translations but the counterpart of the Øverst strings added here: six
dialogs would otherwise show Øverst above Bund.

## AI

each change, before any of it was written.


chrip (approved):
**Approve** — no blockers.

Verified against the merge base: **241 keys added, 29 changed, 0 removed** across 2 file(s). All locale files parse, no placeholder or `<br>` mismatch against the English source, no key-order drift, and every added key exists in `en.json`. Cross-checked against the other 33 open Danish PRs: no key is set to a conflicting value by another branch, so merge order does not matter. CI green, DCO signed.

PR #239, opened 2026-09-18
i18n(da): translate the document editor's forms, messages and protection

bargib:
Two hundred and sixty-four strings: the forms tab and its recipient
dialogs, the messages and error texts, the document protection, the
signature dialog, and the forms and embedded builds. The last of three
pull requests that finish the Danish document editor.

## The error messages open with the fact

Five of them opened alike in English — *An error has occurred while
opening the file* — and then said what was actually wrong. The Danish
style guide asks for the fact first:

```
An error has occurred while opening the file.
   The file content corresponds to text documents (e.g. docx),
   but the file has the inconsistent extension: %1.

Filen kunne ikke åbnes.
   Indholdet er et tekstdokument, fx docx, men filen har endelsen %1.
```

This is the wording settled on for the *failed* family in an earlier
pull request.

One message was turned around. *You are not allowed to edit this
selection because it is protected* opens with a reprimand and ends with
the only useful part. In Danish the reason comes first: **Det markerede
er beskyttet, og du kan derfor ikke redigere det.**

## Two names the source gets wrong

The source calls both the cross shape and the arithmetic sign `Plus`.
They are different shapes, and a gallery listing the same name twice is
not usable, so the signs are **Plustegn** and **Minustegn**.

`Credit Card` is **betalingskort**, not kreditkort. The field takes any
card number, including a Dankort, and a kreditkort is a particular kind
of card.

## Recipient carries the tab

`Recipient` appears twenty-four times and settles the rest:
**Administrer modtagerroller**, *Tilføj modtager*, *Modtagerens navn*.
English has a noun for the person filling a form in — *filler* — that
Danish does not, so those sentences say what the person does instead.

`Radio button` is **radioknap**, in one word. The repository already
used it four times; *alternativknap* came from the Swedish reference
and is not what the rest of this project says.

## Sharing rather than copying

The forms and embedded builds repeat most of these sentences. They now
take the same text as the main build rather than holding their own copy,
so a correction in one place reaches all three.

## AI

each change, before any of it was written.


chrip (approved):
**Approve** — no blockers.

Verified against the merge base: **251 keys added, 13 changed, 0 removed** across 4 file(s). All locale files parse, no placeholder or `<br>` mismatch against the English source, no key-order drift, and every added key exists in `en.json`. Cross-checked against the other 33 open Danish PRs: no key is set to a conflicting value by another branch, so merge order does not matter. CI green, DCO signed.

PR #240, opened 2026-09-19
i18n(da): use one Danish word for each thing, not five

bargib:
Two hundred and seventy-nine strings across all four editors. The first
of three pull requests correcting Danish that was already there.

Nothing here is a new translation. Every string had Danish, and every
one of them was either the wrong word or a fourth spelling of a word the
project had already settled on.

## The scale of the disagreement

A single concept had up to five Danish words:

```
Set          Vælg, Sæt, Angiv, Indstil, Aktiver
border       ramme, grænse, kant
Shape        form (47), figur (27)
sparkline    minidiagram (7), mini diagram, sparkline
Track Changes  Spor ændringer, Spor Ændringer, Sporede ændringer,
               ændringssporingstilstand
```

None of these are close calls. *Vælg* is choose, not set. A *ramme* is a
picture frame. A *fraktion* is a political grouping. A *landskab* is a
painting.

## Where the source itself is ambiguous

`Set` does two different things, and the Danish now follows the
difference rather than the word:

```
Set top border only    Sæt kun øverste kant       a click applies it
Set column width       Angiv kolonnebredde        a value is typed in
Set text language      Vælg tekstsprog            a choice from a list
```

The third is not an oversight. For that one string, German says *wählen*
and Russian *Выбрать* — both choose. The source means choose there.

## How the decisions were checked

Each term was counted across the repository first, then checked against
the established Danish interface, then against German, Swedish and —
where the English could be read two ways — Russian.

Russian earns its place because it cannot hide the difference between a
command and a noun. That mattered for `Group`, which is both in the
source: Сгруппировать is the verb, Группа the noun. Three keys are nouns
and are left alone; one of them builds the string "Gruppe 1" as a name
for a set of radio buttons.

It also caught three places where German is simply wrong and would have
misled us: `Cap type` as *Zierbuchstabe*, `Bet` as *binär*, `Bullet` as
*Zeichen*. Russian is consistent in all three.

## AI

and the reason for each change. Several of my questions changed the
result: *Angiv* was too formal for a button that applies a border, the
shape called a scroll is a rulle rather than a pergamentrulle, and
TextArt is a name that stays in English — written as one word, like
SmartArt.


chrip (approved):
**Approve** — no blockers.

Verified against the merge base: **0 keys added, 294 changed, 0 removed** across 7 file(s). All locale files parse, no placeholder or `<br>` mismatch against the English source, no key-order drift, and every added key exists in `en.json`. Cross-checked against the other 33 open Danish PRs: no key is set to a conflicting value by another branch, so merge order does not matter. CI green, DCO signed.

PR #241, opened 2026-09-19
i18n(da): correct the shape gallery and the arrange row

bargib:
One hundred and fourteen strings: the shape gallery and the ribbon row
that moves shapes around. The second of three pull requests correcting
Danish that was already there.

## Five shapes had become something else entirely

```
Can                Kan              the verb "can"
Chord              Ledning          an electrical cord
Right triangle     Højre trekant    right-hand, not right-angled
Pie                Cirkeldiagram    a chart, not a shape
Diamond            Diamant          the gemstone
```

A user reading the Danish gallery would be looking for a cylinder under
a word that means "is able to".

## The scroll is a scroll, not scrolling

`Horizontal scroll` read **Vandret rul**. The shape is a rolled-up
scroll, the kind that sits in the banners group.

Seven of the eleven languages with a translation made the same mistake —
German *Bildlauf*, Swedish *rullning*, French *Défilement*, Italian
*Scorrimento*, Dutch *scrollen*, Czech *posouvání*, Finnish *skrollaus*.
Russian (*свиток*), Spanish (*Pergamino*) and Polish (*Zwój*) got it.

It is now **Vandret rulle**, without *pergament*: the source says
nothing about the material.

## Two pairs shared one name

`Left bracket` and `Left brace` both read *parantes* — misspelled, and
the same word for ( and {. Russian distinguishes them: круглая скобка is
round, фигурная скобка is curly. So the brackets keep *parentes*, spelt
correctly, and the braces become *tuborgklammer*.

This is worth naming because the obvious fix was wrong. Calling the
bracket a *kantparentes* would have turned a correct translation into an
incorrect one — the shape is ( and not [.

## The arrange row

Four commands had five vocabularies, spread across the editors:

```
Bring Forward        Fremrykke, Ryk frem
Send Backward        Ryk tilbage, Send baglæns, Send tilbage
Bring to foreground  Før til forgrunden, Placer forrest
Send to Background   Flyt til baggrund, Placer bagerst
```

Two of the pairs had already been decided in an earlier pull request,
but only in one editor each. `Align` read *Tilpas*, which is fit rather
than align.

`Group` becomes Gruppér only where it is a command. Three keys are
nouns, and the source cannot show which — but Russian can, and
`Toolbar.js:2454` confirms it by building the string "Gruppe 1".

## AI

and the reason for each change.


chrip (approved):
**Approve** — no blockers.

Verified against the merge base: **0 keys added, 108 changed, 0 removed** across 5 file(s). All locale files parse, no placeholder or `<br>` mismatch against the English source, no key-order drift, and every added key exists in `en.json`. Cross-checked against the other 33 open Danish PRs: no key is set to a conflicting value by another branch, so merge order does not matter. CI green, DCO signed.

PR #242, opened 2026-09-19
i18n(da): fix typos and words that meant something else

bargib:
One hundred and forty strings. The last of three pull requests
correcting Danish that was already there.

Most of these were found by a tool rather than by reading, because they
look like real words.

## The repository is its own dictionary

There is no Danish spell checker available here, but the eleven Danish
locale files hold 3,278 distinct words, and that is a corpus. A word
that sits one edit away from a word eight times more common is almost
always a typo, and that test needs no dictionary at all:

```
okbjekt          2 ×   against  objekt          46 ×
indstilliniger   1 ×            indstillinger  196 ×
indstllinger     3 ×            indstillinger  196 ×
beskytter        2 ×            beskyttet       50 ×
```

*Flyt okbjekt med tekst*. *PDF indstilliniger*. *Arket er beskytter for
redigering*. None of these would survive a careful reading — they simply
were not read, because they look close enough to right.

`formattering` with two t's appeared nine times.

The same method finds run-on words by asking whether the joined form
already exists in the corpus: *tal format* against *talformat*, *linje
afstand* against *linjeafstand*. It reads both ways — *ligmed* appears
three times against *lig med* seventy-one, so there the joined form is
the error.

## Ellipsis is not an ellipse

Seventeen strings confused the character … with the oval shape, which
has its own key in the same file.

## Words that meant something else

| | |
|---|---|
| `Complement` | read **Komplimenter** — compliments. The character is ∁, the set complement, sitting between ∀ and ∂ in `define.js:77`. The key is named `txtSymbol_additional`, which is misleading in a third direction. |
| `Bet` | read **Indsats**, a wager. It is ℶ, the second Hebrew letter. |
| `Scripts` | read **Manuskripter**. It is the group holding superscript and subscript in the equation bar. |
| `Asterisk operator` | read **Stjerne operatør**. An operatør is a person who operates something; the sign is an operator. |

`Text Art` stood in three forms — *Tekst art*, *Textstil*, *Text art*.
It is a feature name and is now **TextArt**, one word, like SmartArt.
Russian keeps it in Latin script inside a Cyrillic sentence — *Вставить
объект Text Art* — which is about as clear a signal as a name can give.

## AI

and the reason for each change.

The measurement tool is in the repository as `vaerktoej/finmasket.py`.
It has four probes and can be pointed at any locale directory.


chrip (approved):
**Approve** — no blockers.

Verified against the merge base: **0 keys added, 140 changed, 0 removed** across 7 file(s). All locale files parse, no placeholder or `<br>` mismatch against the English source, no key-order drift, and every added key exists in `en.json`. Cross-checked against the other 33 open Danish PRs: no key is set to a conflicting value by another branch, so merge order does not matter. CI green, DCO signed.

PR #243, opened 2026-09-20
i18n(da): name the SmartArt layouts and chart options in presentations

bargib:
The SmartArt gallery and the chart panels were the last two galleries in the presentation editor still showing English. 237 strings.

The layout names follow the ones the document editor already uses, so the same layout is called the same thing in both editors. This also resolves a collision: `Pie Process` and `Circle process` had both ended up as "Cirkelproces", and they are now "Cirkeldiagramproces" and "Cirkelproces".

Where a name was missing entirely, it was taken from established Danish terminology rather than translated word for word.

### Consequential changes

Four number formats in the chart axis dialog (10 000 through 100 000 000) are added with the Danish thousands separator.

## AI


chrip (approved):
**Approve** — no blockers.

Verified against the merge base: **237 keys added, 0 changed, 0 removed** across 1 file(s). All locale files parse, no placeholder or `<br>` mismatch against the English source, no key-order drift, and every added key exists in `en.json`. Cross-checked against the other 33 open Danish PRs: no key is set to a conflicting value by another branch, so merge order does not matter. CI green, DCO signed.

PR #244, opened 2026-09-20
i18n(da): translate the presentation menus and the transition panel

bargib:
The right-click menu, the toolbar, the transition and animation panels and the slide master view. 196 strings, plus two rounds of corrections to the slide size dialog.

Moving a slide is now "Flyt slide til slutningen" and "Flyt slide til begyndelsen". The earlier wording mixed two systems, and "forrest" is needed for Bring to front, which is about layering rather than order.

### Consequential changes

The theme colour tooltip in the palette says Fremhævning, matching the three other editors, which already did.

The slide size dialog had a set of errors that showed up while testing this branch:

- The widescreen size was called Bredskærm in the dialog and Widescreen   in the menu. The dialog builds three entries from that one string by   appending the ratio, so the two names sat next to each other in the   same list. It is Widescreen in both places now.
- The letter size read "8,5 x 11 i" - the unit had lost its second   letter and said nothing. It stays as `in`, matching Ledgerpapir, which   was already correct.
- The paper names were written as two words with a capital in the middle,   and A3 had no space before its measurement.

Two strings in that dialog are still wrong and are deliberately left alone: Overhead reads "Overliggende", which means something else, and Custom is in the plural. Both are touched by open PRs that fix only their capitalisation, so correcting the words here would conflict.

## AI


chrip (approved):
**Approve** — no blockers.

Verified against the merge base: **179 keys added, 21 changed, 0 removed** across 1 file(s). All locale files parse, no placeholder or `<br>` mismatch against the English source, no key-order drift, and every added key exists in `en.json`. Cross-checked against the other 33 open Danish PRs: no key is set to a conflicting value by another branch, so merge order does not matter. CI green, DCO signed.

PR #245, opened 2026-09-20
i18n(da): describe the keyboard shortcuts in the presentation editor

bargib:
68 shortcut descriptions for the presentation editor.

These are written from the behaviour of the editor rather than translated word for word. The English text often describes two different outcomes in one sentence, and Danish reads better when the condition comes first: 
"Er hele tabellen markeret, flyttes markøren til den første celle" rather than a literal rendering of "If the table is selected, move the cursor to the first cell".

The shortcut labels themselves are left untouched. They have no English source text to translate from - the English file repeats the key name.

## AI


chrip (approved):
**Approve** — no blockers.

Verified against the merge base: **68 keys added, 0 changed, 0 removed** across 1 file(s). All locale files parse, no placeholder or `<br>` mismatch against the English source, no key-order drift, and every added key exists in `en.json`. Cross-checked against the other 33 open Danish PRs: no key is set to a conflicting value by another branch, so merge order does not matter. CI green, DCO signed.

PR #246, opened 2026-09-20
i18n(da): translate the presentation dialogs, messages and settings

bargib:
The remaining dialogs, the settings panels, the presenter view, the print dialog and the error and status messages, plus the embedded viewer and a few mobile strings. 212 strings.

This is the last batch for the presentation editor. Together with the three sibling PRs it finishes this editor.

### Consequential changes

The error messages follow the pattern already agreed for the other editors: a full sentence with a definite article, so "Filen kunne ikke indlæses" rather than a bare fragment.

`PE.Views.HeaderFooterDialog.textSlide` is deliberately left out. It is already in #235 with exactly the same text, and changing it in two places would only create noise.

So after merging and previous PRs I have a few fixes to run through.

## AI


chrip (approved):
**Approve** — no blockers.

Verified against the merge base: **194 keys added, 18 changed, 0 removed** across 3 file(s). All locale files parse, no placeholder or `<br>` mismatch against the English source, no key-order drift, and every added key exists in `en.json`. Cross-checked against the other 33 open Danish PRs: no key is set to a conflicting value by another branch, so merge order does not matter. CI green, DCO signed.

PR #248, opened 2026-09-20
i18n(da): translate the spreadsheet toolbar and the right-click menu

bargib:
The toolbar, the right-click menu, the view and formula tabs and the
table design tab. 230 strings.

Two names were decided against the majority in the repository:

- **Collapse** becomes "Fold sammen" rather than "Kollaps", which is not
  Danish in that sense. Not "Skjul" either, since Hide sits right next to
  it in the same menu and they are different commands.
- **Count** becomes "Antal" rather than "Tæl". Its neighbours Sum,
  Gennemsnit and Produkt are all nouns.

The symbol names follow one pattern throughout: an English "X sign"
becomes a Danish "X-tegn".

## AI

Used Claude Opus 5 to assist in making this PR. All translations have been verified
by myself prior to pushing them.


chrip (approved):
**Approve** — no blockers.

Verified against the merge base: **210 keys added, 20 changed, 0 removed** across 1 file(s). All locale files parse, no placeholder or `<br>` mismatch against the English source, no key-order drift, and every added key exists in `en.json`. Cross-checked against the other 33 open Danish PRs: no key is set to a conflicting value by another branch, so merge order does not matter. CI green, DCO signed.

PR #254, opened 2026-09-20
i18n(da): translate the solver and goal seek

bargib:
110 strings for the solver and goal seek, neither of which had any
Danish.

The terminology is taken from established Danish usage for this kind of
dialog rather than invented: the objective is a *målsætning*, the
*Evolutionary* method is *Udvikling*, and *Simplex LP* keeps its name.

Coverage of the spreadsheet editor: 73.9% -> 75.6%.

### Worth a second opinion

- **`Solving limits`** is rendered *Grænser for løsningen*. The
  conventional Danish for this group reads awkwardly, so this departs
  from it deliberately.
- **`int`, `bin` and `dif`** in the constraint dropdown stay as they are.
  The full Danish word for *int* is *heltal*; I could not confirm the
  full forms of *bin* and *dif*, so they are left short.
- **The solver is a name, not a common noun** here: *Problemløser fandt
  en løsning*, without the definite article.

### Conflicts

None. Every key in this branch was compared against all my other open
branches. Across the whole set only five keys appear in more than one
branch, and all five carry identical text.

### Verification

Smoke-tested in a container built from main with this branch applied.

## AI

Used Claude Opus 5 to assist in making this PR: to collect the
untranslated strings, to cross-check wording against the other languages
in this repository, and to draft the Danish.

Every string in this branch has been reviewed by me individually, one at
a time, from a sheet listing the English source, the renderings in the
other languages, the existing Danish where there was one, and the
proposed text. Nothing here went in unread. The wording is mine, I have
verified it against the running editor, and I stand behind the whole
translation and take responsibility for it.


chrip (approved):
**Approve** — no blockers.

Verified against the merge base: **110 keys added, 0 changed, 0 removed** across 1 file(s). All locale files parse, no placeholder or `<br>` mismatch against the English source, no key-order drift, and every added key exists in `en.json`. Cross-checked against the other 33 open Danish PRs: no key is set to a conflicting value by another branch, so merge order does not matter. CI green, DCO signed.

PR #259, opened 2026-09-20
i18n(da): translate redaction, form fields and number formats in PDF

bargib:
197 strings: the redaction tab, the form field settings and the number
formats. The redaction tab had no Danish at all.

Coverage of the PDF editor: 39.6% -> 46.3%.

### Worth a second opinion

- **Redact is "overstregning".** That is the word used when part of a
  document must not be readable. German says "schwärzen", French
  "biffer". "Rediger" cannot be used here: the menu has Edit Text and
  Redact Text side by side.
- **Social Security Number reads "Personnummer", not CPR-nummer.** The
  field validates an American number format, so CPR would promise
  something it cannot keep. German uses its own national term the same
  way.
- **Rollover is "Musen over".** German reads it as "turning around" and
  Inset as "deployment"; Russian has both right.

### Conflicts

None. Every key was compared against my other open branches, key by
key: no key appears in two of the six, and none disagrees with the
other open branches.

### Verification

Smoke-tested in a container built from main with all six branches
applied. All 1,559 strings were present and correct.

## AI

Used Claude Opus 5 to assist in making this PR: to collect the
untranslated strings, to cross-check wording against the other languages
in this repository, and to draft the Danish.

Every string in this branch has been reviewed by me individually, one at
a time, from a sheet listing the English source, the renderings in the
other languages, the existing Danish where there was one, and the
proposed text. Nothing here went in unread. The wording is mine, I have
verified it against the running editor, and I stand behind the whole
translation and take responsibility for it.


chrip (approved):
**Approve** — no blockers.

Verified against the merge base: **197 keys added, 0 changed, 0 removed** across 1 file(s). All locale files parse, no placeholder or `<br>` mismatch against the English source, no key-order drift, and every added key exists in `en.json`. Cross-checked against the other 33 open Danish PRs: no key is set to a conflicting value by another branch, so merge order does not matter. CI green, DCO signed.

